PR #7, opened 2017-09-25
Add support for Logitech G430 on macOS

joeslazaro:
This code works with the full range of volume settings, tested on macOS 10.13 (High Sierra).

It fails on Linux with the following error: 
  libusb: error [submit_control_transfer] submiturb failed error -1 errno=16


Sapd:
Thank you for your contribution.

Could be that there is some problem with LIBUSB_REQUEST_CLEAR_FEATURE on linux, maybe it's not the right one. While sniffing I noticed that some tools report differently then others, there could lie the error cause.

PR #10, opened 2018-02-27
Adding G533 Support

LowlySysadmin:
Tested this with various different volume levels on Ubuntu 17.10 with Logitech G533.  Seems to work great.


Sapd:
Thanks, I merged it into master.

Issue #13, opened 2018-03-27
VOID PRO battery check issue


kuchenstrudel:
When the mic is put up vertically, which triggers an automatic mute, the battery readings are wrong.

Examples: 

> mic down, muted (with the mute key)
> Battery: 94%

> mic down, unmuted
> Battery: 94%

> mic up, automuted
> Battery: 222%

Maybe this is an issue with the VOID PRO and you are not getting it with yours, if so: can I provide you with any readings to figure this out?

Sapd:
Could be a simple flag/bitmask, because 222-94=128. You could try it again with a different battery level, check the level, put the mic up and check the level again and subtract 128 to see if the result is the same.

kuchenstrudel:
Unplugged the headset, and spammed `HeadsetControl -b`

(mic up)
> Battery: 222%
 
> Failed to request battery. Error: -100: (null)

> Battery: 223%

> Battery: 223%
 
> Battery: 223%

(mic down)
> Battery: 95%

223-95 -> 128, will leave it unplugged for a while and report back once again.

kuchenstrudel:
(mic up)
> Battery: 220%

(mic down)
> Battery: 92%

220-92 -> 128, seems you are correct.

Sapd:
Then you can do something like this:
```c
// on top of the c file
enum voidpro_battery_flags {
     VOIDPRO_BATTERY_MICUP = 128
}

// then inside the function
if (data_read[4] & VOIDPRO_BATTERY_MICUP)
    return data_read[4]-VOIDPRO_BATTERY_MICUP;
```

This will simply return the battery status. We could also pass it, however then the battery_status enum has to be updated.

kuchenstrudel:
I adjusted your code a bit and it looks good (as in working) now. See https://github.com/Sapd/HeadsetControl/pull/14

kuchenstrudel:
Fixed in https://github.com/Sapd/HeadsetControl/commit/5f49bd63d257af57e348760e00e2fe1b21077d7a, closing.

PR #14, opened 2018-03-27
Fixes issue #13

kuchenstrudel:
Works for me.


Sapd:
I would suggest changing the modulo in the `if` to `&` as to make clear that a bitmask is used. (In fact I never saw modulo used in software as replacement for bit operators).
Probably we should even use `return data_read[2] &~ VOIDPRO_BATTERY_MICUP;`. But that doesn't really matter.

Sapd:
Thanks, I added some comments and updated code style (when parentheses are required, every corresponding scope on the same level also shall use parentheses).

Issue #8, opened 2017-09-27
Logitech G430 on Linux


joeslazaro:
I tried again to get this working on Linux, and I discovered additional information dmesg output when I ran headsetcontrol:
```
[523380.230450] usb 3-9: usbfs: process 6312 (headsetcontrol) did not claim interface 0 before use
[523380.232022] input: Logitech Logitech G430 Gaming Headset as /devices/pci0000:00/0000:00:14.0/usb3/3-9/3-9:1.3/0003:046D:0A4D.0010/input/input27
[523380.290414] hid-generic 0003:046D:0A4D.0010: input,hidraw1: USB HID v1.00 Device [Logitech Logitech G430 Gaming Headset] on usb-0000:00:14.0-9/input3
```

If I change the calls to libusb_claim_interface() and libusb_release_interface() to use 0 for the second argument, then it works.

I don't want to submit a PR with that change because I assume that it would break everything that was already working. Do you think there is a way to dynamically discover the correct value to use depending on device and platform?

joeslazaro:
I looked further at the lsusb output, and found that the interface number that I needed (0) corresponds to the entry where bInterfaceClass = 1 Audio && bInterfaceSubClass = 1 Control Device
```
joe@ironforge:~/Projects/HeadsetControl-g430$ lsusb -d 046d: -v

Bus 003 Device 008: ID 046d:0a4d Logitech, Inc. G430 Surround Sound Gaming Headset
Device Descriptor:
  bLength                18
  bDescriptorType         1
  bcdUSB               1.10
  bDeviceClass            0 (Defined at Interface level)
  bDeviceSubClass         0 
  bDeviceProtocol         0 
  bMaxPacketSize0        16
  idVendor           0x046d Logitech, Inc.
  idProduct          0x0a4d G430 Surround Sound Gaming Headset
  bcdDevice            1.20
  iManufacturer           1 Logitech
  iProduct                2 Logitech G430 Gaming Headset
  iSerial                 0 
  bNumConfigurations      1
  Configuration Descriptor:
    bLength                 9
    bDescriptorType         2
    wTotalLength          253
    bNumInterfaces          4
    bConfigurationValue     1
    iConfiguration          0 
    bmAttributes         0x80
      (Bus Powered)
    MaxPower              100mA
    Interface Descriptor:
      bLength                 9
      bDescriptorType         4
      bInterfaceNumber        0
      bAlternateSetting       0
      bNumEndpoints           0
[...]

PR #16, opened 2018-04-19
Add support for SteelSeries Arctis 7.

Ionic:
This currently only supports setting (or disabling) the sidetone feature.

Sadly, this control station falls out of line by using Output Requests instead of Feature Requests, so the code is a tad different from all the other implementations.

Further, the request needs to be sent to a non-default interface, which touches on #8. Thus, I've implemented initial interface opening, which SHOULD be backwards compatible due to initializing the device structure statically (and thus nulling the memory block) and assuming that `hid_enumerate` returns the default interface as the first one - which would be what `hid_open(vid, pid)` would normally open.

I was not able to actually test this, though, so please go ahead and make sure it doesn't break your devices.

### DIFF
--- a/src/devices/steelseries_arctis7.c
+++ b/src/devices/steelseries_arctis7.c
@@ -1,0 +1,57 @@
+#include "../device.h"
+#include "../utility.h"
+
+#include <hidapi.h>
+#include <string.h>
+#include <stdlib.h>
+
+static struct device device_arctis7;
+
+static int arctis7_send_sidetone(hid_device *device_handle, uint8_t num);
+
+void arctis7_init(struct device** device)
+{
+    device_arctis7.idVendor = VENDOR_STEELSERIES;
+    device_arctis7.idProduct = 0x1260;
+    device_arctis7.idInterface = 0x05;
+
+    strcpy(device_arctis7.device_name, "SteelSeries Arctis 7");
+
+    device_arctis7.capabilities = CAP_SIDETONE;
+    device_arctis7.send_sidetone = &arctis7_send_sidetone;
+
+    *device = &device_arctis7;
+}
+
+static int arctis7_send_sidetone(hid_device *device_handle, uint8_t num)
+{
+    int ret = -1;
+
+    // the range of the Arctis 7 seems to be from 0 to 0x12 (18)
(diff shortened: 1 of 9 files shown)


src/devices/steelseries_arctis7.c
```diff
@@ -0,0 +1,51 @@
...
+
+    *device = &device_arctis7;
+}
+
+static int arctis7_send_sidetone(hid_device *device_handle, uint8_t num)
+{
+    // the range of the Arctis 7 seems to be from 0 to 0x12 (18)
+    num = map(num, 0, 128, 0x00, 0x12);
+
+    unsigned char *buf = calloc(31, 1);
```
Sapd: buf is not getting free'd

src/devices/steelseries_arctis7.c
```diff
@@ -0,0 +1,51 @@
...
+    {
+        return -1;
+    }
+
+    const unsigned char data_on[5] = {0x06, 0x35, 0x01, 0x00, num};
+    const unsigned char data_off[2] = {0x06, 0x35};
+
+    if (num)
+    {
+        memmove(buf, data_on, sizeof(data_on));
```
Sapd: memmove is usually used when there is overlapping memory, normally I would use memcpy, but it doesn't really matter much

Sapd:
Also I tested wether it still works with my Headset, there doesn't seem to be any problem and I also guess that there won't be with other ones. Otherwise the interface id's of the other headsets must be adapted.

src/devices/steelseries_arctis7.c
Ionic: Good catch.

Regarding `memcpy` vs. `memmove`: I try to avoid `memcpy` since its potential benefit (increased speed potentially through prefetching) outweigh the risk (not safe with overlapping regions). Especially in this case it won't make any difference, neither runtime-related nor behavioral.

Ionic:
Updated the branch. Some more cleanup, less code duplication etc.

Ionic:
Err... I don't get why git sometimes changes more commits than would be needed, so threw in a quick branch rebase.

Sapd:
Thank you, I merged it into master

PR #24, opened 2018-07-06
Changed CMake to prefer hidapi-hidraw over hidapi-libusb.

### LINKED ISSUE #23: volume scroll stops working after launching
Device: G533

This works before launching HeadsetControl
The headset itself makes a little sound indicating if you are raising or lowering the volume, volume raises or lowers fine and in my case, mate desktop shows this notification:
gi

After launching `sudo HeadsetControl -s 128`
you can still hear the sound in headset, but volume remains unchanged and no notification is shown, off course.

edit: running in debian stretch with mate.

Eusebius1920:
Fixes #23


Sapd:
Thank you

alete89:
thanks both

PR #30, opened 2018-11-20
Arch Linux dependencies.

Fiiily:
These dependencies seemed to be what I needed to get things going on my Antergos (arch-based) system.


Sapd:
Aren't compilers missing, or does arch linux require them anyways?

parkerlreed:
It's assumed you have base-devel installed (which incluides gcc/make/etc)

PR #32, opened 2018-12-02
Rework udev-rules

### LINKED ISSUE #31: hidraw-Permissions
I'm on Arch and had the Problem, that i needed to be root to use HeadsetCotrol on my corsair void Pro.
The reason for this is, afaik, that the hidraw-devices used by this program are owned by root. To change this I changed the udev-rule to contain the Following:

`SUBSYSTEM=="hidraw", SUBSYSTEMS=="usb", ATTRS{idVendor}=="1b1c", ATTRS{idProduct}=="0a14", MODE="0666"`

With the `ENV{ID_MODEL}="..."` my headset did not get matched, and as the Vendor-/Product-ID do not change it is afe to use them. Further changes to `SUBSYSTEM` were made to change the Permissions for the correct `/dev/hidraw#`.

As I am not sure whether this is specific for Archlinux or applicable on other Distributions, i would ask you to look into this and if necessary change the rules.

Regards
Hades

Hades0299:
Update the udev-rules to hidraw (#31) and rename the files to match the naming-convention. Further added a g533-rule.


Sapd:
Thank you

PR #33, opened 2018-12-06
Added a switch for short output: -c

Hades0299:
For usage in scripts I do not want to parse the output and filter out
the usefull infomration or clutter the line with redirects to /dev/null.
Therefore I added a option, that removes every output, except whith the
-b-switch, where the battery-percentage (or -1 for loading) are printed.

While doing so, the default changed a bit:
Only when the -h-swich is present, we output name and author of this
programm.

Further: my Texteditor removed some whitespaces in empty lines....

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -27,9 +27,14 @@
 #include <unistd.h>
 #include <string.h>
 
-
 // describes the device, when a headset was found
 static struct device device_found;
+
+// 0=false; 1=true
+static int short_output = 0;
+
+#define PRINT(command) { if (!short_output) \
+    {command;}}
 
 /**
  *  This function iterates through all HID devices.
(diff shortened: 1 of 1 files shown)


Sapd:
Looks good so far. Any reason the PRINT macro doesn't call printf directly?

Hades0299:
The only reason the macro does not call printf is, that i wrote this while i was quite tired, and couldn't figure out how to pass multiple arguments through a macro to printf. I can try to figure that out, but if you have an Idea, I'm all ears ;)

Sapd:
Macros support variadic arguments https://stackoverflow.com/questions/20639632/how-to-wrap-printf-into-a-function-or-macro

Hades0299:
Seems that my abilities to use google need updating.....

Sapd:
Looks good, thanks for your contribution.

PR #39, opened 2019-08-04
Support switching LED off/on for Corsair Void Pro

felfert:
This PR adds support for switching LED lights off/on for Corsair Void Pro.
Switching LEDs off almost doubles the battery lifespan.

While adding CAP_LIGHTS, I also noticed and fixed a bug:
Since capabilities are a bitmap, CAP_NOTIFICATION_SOUND should be 4, not 3!

Cheers
 -Fritz


Sapd:
Thanks! I also verified it with the normal corsair void and it works accordingly too.

Issue #35, opened 2019-01-28
Steelseries Artic 7 2019 not reconnized


olielvewen:
Hi,
I've purchased for Christmas an Steelseries Headset Artic 7 version 2019 and it is not reconnized on Linux Manjaro with the last version of HeadsetControl.
<code>[olivier@thorielle ~]$ HeadsetControl -h
Headsetcontrol written by Sapd (Denis Arnst)
        https://github.com/Sapd

Parameters
  -s level      Sets sidetone, level must be between 0 and 128
  -b            Checks the battery level
  -n soundid    Makes the headset play a notifiation

[olivier@thorielle ~]$ HeadsetControl -b
Headsetcontrol written by Sapd (Denis Arnst)
        https://github.com/Sapd

No supported headset found
[olivier@thorielle ~]$ HeadsetControl -n 0
Headsetcontrol written by Sapd (Denis Arnst)
        https://github.com/Sapd

No supported headset found</code>

Here more explanation about my system and usb:

<code>[olivier@thorielle ~]$ lsusb
Bus 004 Device 001: ID 1d6b:0003 Linux Foundation 3.0 root hub
Bus 003 Device 002: ID 046d:c517 Logitech, Inc. LX710 Cordless Desktop Laser
Bus 003 Device 001: ID 1d6b:0002 Linux Foundation 2.0 root hub
Bus 002 Device 001: ID 1d6b:0003 Linux Foundation 3.0 root hub
Bus 001 Device 004: ID 1b1c:1b51 Corsair 
Bus 001 Device 003: ID 046d:c52b Logitech, Inc. Unifying Receiver
Bus 001 Device 002: ID 1038:12ad SteelSeries ApS 
Bus 001 Device 001: ID 1d6b:0002 Linux Foundation 2.0 root hub

<code>
[olivier@thorielle ~]$ dmesg | grep usb
[    0.718696] usbcore: registered new interface driver usbfs
[    0.718696] usbcore: registered new interface driver hub
[    0.718700] usbcore: registered new device driver usb
[    0.869704] usbcore: registered new interface driver usbserial_generic
[    0.869708] usbserial: USB Serial support registered for generic
[    1.718711] usb usb1: New USB device found, idVendor=1d6b, idProduct=0002, bcdDevice= 4.19
[    1.718712] usb usb1: New USB device strings: Mfr=3, Product=2, SerialNumber=1
[    1.718713] usb usb1: Product: xHCI Host Controller
[...]

PR #40, opened 2019-08-04
SteelSeries Arctis 7 Battery Support + 2019 Model Support

jamesj999:
Added battery support and support for 2019 variant of SteelSeries Arctis 7.


jamesj999:
Fixes https://github.com/Sapd/HeadsetControl/issues/35

Sapd:
Thank you!
Is the Request still Work-In-Progress? (As some commits were added some hours ago).
Also the README needs an update, I can also take care of it, though.

jamesj999:
Hello there!

Apologies, we were slightly premature with the request - we spotted a couple of issues and fixed them. We're just double checking the fully charged value as I've seen it go over 100% so we're verifying on a second headset and will cap it at 100% if necessary. I'll update the readme now and will let you know when we've checked the battery percentage issue.

Thanks,
James

raiscan:
> Thank you!
Thank you! We had a good basis to get our headset battery in Linux thanks to this project.

raiscan:
After experimenting, I believe their battery percentage response isn't calibrated very well or calibrated to a reference battery; mine never goes above 99, while @jamesj999 's can reach all the way to 116 before capping out. The SteelSeries UI in Windows uses a simple battery segment indicator so I imagine they're aware it's not particularly accurate.

When I get home tonight I'll just limit the max output to 100. It'll be good enough for anyone's purposes; especially since unplugging James's heatset immediately drops it lower than 100 :)

Sapd:
Yes I think this is not uncommon, there are two ways to measure how full a battery is.

The first (halfway accurate) method is to empty the battery completely and charge it until the maximum voltage is reached and count the watt-hours. (Now you can simply integrate the watt-hours while using the headset and know how much of the maximum number is left).

The second method (probably with most headsets) is to simply measure the voltage. It's very inaccurate and not really linear. Therefore, it may well be that the value is also inaccurate.

raiscan:
Hello! I forgot last night. I've made some changes to the arctis 7 rules due to the old udev rules not working (See commits 669aac0 & 2128b78). I have also capped the battery readout on 100% to both arctis 7 and arctis 7 2019 as they seem to be pretty darn similar and we've seen it on one at least.

Please let me know if there's anything you'd like us to change, but otherwise I think we're done :) Thanks!

Sapd:
Looks good, maybe the UDEV rules need to be adjusted for other devices, but this pull request is fine.

PR #44, opened 2019-08-20
Added Support Logitech G933

KrDimitrov:
# Support G933
Added partial support for the Logitech G933 Wireless. 

Currently, only the battery status report feature is implemented. Sadly, **its only an estimate**. The battery voltage is mapped to levels, thanks to https://github.com/ashkitten/g933-utils , I got the map from there. 

Since battery voltage fluctuates, the reported battery level will too fluctuate by about 3%. 

If the Headset is OFF, e.g the USB adapter is plugged in, but the headset is not connected to it, the battery level should report 0%.

I will try and implement more features like LED, sidetone, etc. Would appreciate some help, and/or code review, since I'm not that good with C.


KrDimitrov:
# Sidetone
Setting sidetone is now possible thanks to the awesome project [notes](https://github.com/ashkitten/g933-utils/tree/master/notes)

**Sidetone level is capped at 100**. Setting it more than 100 will send the same request as at level 100.

Looks like lighting configuration would be possible to implement, with custom lighting support, and preset effects as well!

Sapd:
Also I noticed, the G633 has the same packet of Sidetone.
https://github.com/Sapd/HeadsetControl/blob/master/src/devices/logitech_g633.c

So probably it has completely the same protocol and we could merge those together.

KrDimitrov:
> So probably it has completely the same protocol and we could merge those together.

Could indeed be, but the G633 is not wireless and has no battery. How could we merge them and keep the battery feature? I also don't have a g633, for testing...

CMakeLists.txt
```diff
@@ -26,6 +26,7 @@ add_subdirectory(src/devices)
 
 add_executable(HeadsetControl ${SOURCE_FILES})
 target_link_libraries(HeadsetControl ${HIDAPI_LIBRARIES})
+target_link_libraries(HeadsetControl m)
```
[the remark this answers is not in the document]
KrDimitrov: I needed some math functions and I read that's how one links the math library, but since we are going to use a function, it's not needed anymore

Sapd:
> but the G633 is not wireless and has no battery. 

Okay, then that's a little disadvantageous. I'm not quite sure myself how we do it the cleanest way.
When you look into corsair_void.c, you see that it's done there, it simply holds a list of multiple device ids. So we can match multiple devices to one source-code.
One way would be to check the device id in g933_request_battery again, and exclude those (return something negative), which don't have a battery. But I would suggest to leave it as it is and have two separate implementations for now.

Regarding LEDs:
When you look into the implementation, you see that the project has general support for switching light on/off. That is easy implementable.
Problem with other settings is, that they are very specific for each device (Like light-transitions, supported light modes etc). So we can't provide general parameters or function headers for them.
Implementing that would probably require a very different approach. Therefore, it is best to do this in a separate issue/PR.

KrDimitrov:
Sorry for the late reply, I was on vacation...

> f(x)=3.7268473047*10^(-9)x^(4)-0.00005605626214573775*x^(3)+0.3156051902814949*x^(2)-788.0937250298629*x+736315.3077118985

I can't get this function to work. It seems like its higher by 5 than the actual value, e.g at voltage 4057 it returns 105.43.
I don't know how you converted it to a function. Should I just subtract, like 5 from it?

We would also lose some precision, and my implementation is already inaccurate.

Sapd:
> Sorry for the late reply, I was on vacation...
> 
> > f(x)=3.7268473047_10^(-9)x^(4)-0.00005605626214573775_x^(3)+0.3156051902814949_x^(2)-788.0937250298629_x+736315.3077118985
> 
> I can't get this function to work. It seems like its higher by 5 than the actual value, e.g at voltage 4057 it returns 105.43.
> I don't know how you converted it to a function. Should I just subtract, like 5 from it?
> 
> We would also lose some precision, and my implementation is already inaccurate.

Such curves are very inaccurate anyway. That is the problem if you only use voltage. A more precise method is to integrate the current flow (in Amperhours, like done in smartphones).
That's why I don't know exactly why the other project provides so many data points. These are only exactly tailored to the battery the data points were generated with. Even if you have the same headset with the same battery model, it can be very different for you. This even changes with the age and usage of the battery.
Practically, with the method you are able to completely roughly say how full it is, I would say approximately to a precision of 1/5.

So basically with x > 4030 you can simply report it as 100.

KrDimitrov:
Oh, I understand now. Thanks for the explenation. Well, I'll try to close this PR in the next couple of days!

CMakeLists.txt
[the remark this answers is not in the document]
KrDimitrov: The m option would be needed again if I implement your function using `pow`. Is that okay?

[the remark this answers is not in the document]
Sapd: Yeah if it doesn't compile without the option.
You can btw. write it then int one line:
 `target_link_libraries(HeadsetControl m ${HIDAPI_LIBRARIES})`

KrDimitrov:
# Light Setting
Light setting is now supported. Setting the light to ON will enable the 'breathing' LED effect on the headset. This is currently not configurable, since this software only allows ON and OFF settings. 

Im not sure if the way I implemented this is good, so please, let me know!

src/device_registry.c
```diff
@@ -21,6 +22,7 @@ void init_devices()
     g533_init(&devicelist[2]);
     g633_init(&devicelist[3]);
     g930_init(&devicelist[4]);
+    g933_init(&devicelist[6]);
     arctis_init(&devicelist[5]);
```
Sapd: Also you need to swap the numbers then, because when someone else adds another device, he can easily miscount it.

src/devices/logitech_g933.c
```diff
@@ -0,0 +1,108 @@
...
+    #endif
+    // actual voltage is byte 4 and byte 5 combined together
+    const uint16_t voltage = (data_read[4] << 8) | data_read[5];
+    
+    #ifdef DEBUG
+    printf("Reported Voltage: %2f\n", (float)voltage);
+    #endif
+
+    return round(estimate_battery_level(voltage));
+
```
Sapd: Remove whitespace

src/devices/logitech_g933.c
```diff
@@ -0,0 +1,108 @@
...
+    }
```
Sapd: The brace and the start of g33_send_sidetone is misaligned

src/devices/logitech_g933.c
```diff
@@ -0,0 +1,108 @@
+#include "../device.h"
+
+#include <string.h>
+#include <math.h>
+#include <stdlib.h>
+#include <hidapi.h>
+//#define DEBUG
```
Sapd: Remove the comment, it can be defined via command-line option while compiling

Sapd:
I squashed & merged it here https://github.com/Sapd/HeadsetControl/commit/38eb1341bda1993ba67f745f07d8f7db58f5455b and fixed a minor compilation warning
(GitHub does not seem to automatically recognise squashing via command line)

PR #50, opened 2019-10-29
Added Support for G935

gort818:


Sapd:
Instead of creating new files for G935, you can simply rename the one of G933 to G93X and add the device id to the array on line 21 (like done in corsair_void.c)

Sapd:
The request for the sidetone differs in the implementation. So we can't be sure that the implementation of G933 will work for that too. 
I also don't understand why so many Logitech devices differ in protocol, whereas Corsair always seem to use the same.

gort818:
That is what I am saying if we use G93X .. that would mean all  G93X devices which could be confused as G930.

But I can do that no problem if that works for you.

Sapd:
Ah didn't thought about that, another suggestion is simply logitech_g933_935.c

gort818:
Ok check it out

src/device.h
```diff
@@ -20,7 +20,7 @@ enum capabilities {
 /** @brief Flags for battery status
  */
 enum battery_status {
-    BATTERY_LOADING         = 65535
+    BATTERY_CHARGING         = 65535
```
Sapd: Such changes are better done in a seperate commit/PR

src/devices/logitech_g933_935.c
```diff
@@ -5,31 +5,35 @@
 #include <stdlib.h>
 #include <hidapi.h>
 
-static struct device device_g933;
+static struct device device_void;
```
Sapd: Note you named it "void"

src/devices/logitech_g933_935.c
```diff
@@ -5,31 +5,35 @@
...
+static const uint16_t PRODUCT_IDS[] = {ID_LOGITECH_G933,ID_LOGITECH_G935};
 
-void g933_init(struct device ** device)
-{ 
-    device_g933.idVendor = VENDOR_LOGITECH;
-    device_g933.idProductsSupported = &PRODUCT_ID;
-    device_g933.numIdProducts = 1;
+static int void_send_sidetone(hid_device *device_handle, uint8_t num);
+static int void_request_battery(hid_device *device_handle);
+static int void_lights(hid_device *device_handle, uint8_t on);
```
Sapd: Also here you named it void

src/device.h
Sapd: Ah I just saw that it's done seperatly and the rest is squashed. Then it's okay in this case

gort818:
oops. fixed

PR #60, opened 2019-12-30
Corsair Void (pro) should return HSC_ERROR when disconnected

olback:
Return `HSC_ERROR` when the headset is disconnected, just like it's done for the [Logitech g533](https://github.com/Sapd/HeadsetControl/blob/master/src/devices/logitech_g533.c#L67).

Alternatively, add a new `DISCONNECTED` state?

### DIFF
--- a/src/device.h
+++ b/src/device.h
@@ -72,7 +72,7 @@ struct device
      *                          device as defined here (same ids)
      *
      *  @returns    >= 0            battery level (in %)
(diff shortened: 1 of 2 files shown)


Sapd:
I think it would be better to add a new state for that and make an error accordingly.

Note that I'm doing some changes regarding syntax format d527e70405862be84e7529d0515f5c867bb1612d To avoid spread out syntax changes. So it is a good idea to revert purely syntax changes in your pull request (for now).

PR #62, opened 2020-01-02
UsageID support and Protocol corrections

Sapd:
Depends on https://github.com/Sapd/HeadsetControl/pull/61

- Implements Windows support, by considering usageID while searching for headsets. UsageIDs are deactivated for other platforms
- The implementations need to send the correct amount of bytes, as Windows is more strict. I corrected the amount of bytes where I know the correct one. This will also be done for corsair void. However I don't know the correct amount for Steelseries and the battery of G533.

- [x]  Correct implementations for G930, 6633/G933/G935, G533(Sidetone)
- [x]  Correct implementation for Corsair void
- [ ]  Correct implementation for G533 (Battery)
- [ ]  Correct implementation for Steelseries


Sapd:
Implementing the correct Windows version for Corsair is not so easy... it has different endpoints for sidetone and battery. However the software just supports one for now. I set the correct one for sidetone.

Issue #66, opened 2020-01-15
Logitech G533 Battery Level


sistematico:
```
[lucas@majestic ~]:$ sudo headsetcontrol -b
Found Logitech G533!

Failed to request battery. Error: -100: (null)
```
Sidetone works ok.
Any help is welcome.

Sapd:
-100 accords to  (we should be probably more specifically tell the error)
```
   //Headset offline
    if (data_read[2] == 0xFF)
        return HSC_ERROR
```

Can you try restarting the headset and plugging usb out and in?

If that does not help, comment the lines 78 and 79 in src/devices/logitech_g533.c out and report back.

sistematico:
@Sapd thank for your awesome help.

Un-plugging and plugging the -b works, but if -s is used show me the new error(with the lines commented):

`Failed to request battery. Error: -48: (null)`

I'm using Arch Linux and my HeadSet have 1-day of use

sistematico:
![Captura de tela_2020-01-15_18-08-33](https://user-images.githubusercontent.com/97509/72475804-0d078080-37c2-11ea-89b4-a22d493ef6da.png)

Sapd:
What happens when you set `MSG_SIZE` on line 31 to 12 and recompile? (I changed it recently).

Otherwise @giuseppeM99 implemented the device, I can't be of any more help as I don't own it.

sistematico:
```
[lucas@majestic build]:$ cmake ..
-- The C compiler identification is GNU 9.2.0
-- The CXX compiler identification is GNU 9.2.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found HIDAPI: /usr/lib/libhidapi-hidraw.so  
-- Configuring done
-- Generating done
-- Build files have been written to: /home/lucas/github/HeadsetControl/build
[lucas@majestic build]:$ make 
Scanning dependencies of target headsetcontrol
[...]

PR #67, opened 2020-01-18
G533 fix

giuseppeM99:
Fix issue #66 on Logitech G533

### DIFF
--- a/src/devices/logitech_g533.c
+++ b/src/devices/logitech_g533.c
@@ -1,5 +1,6 @@
 #include "../device.h"
 #include "../utility.h"
+#include "logitech.h"
 
 #include <math.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


Sapd:
In the first commit you used MSG_SIZE (being 64 bytes) and in the second commit HIDPP_LONG_MESSAGE_LENGTH (which is 20 bytes). Is this intended?

giuseppeM99:
Yes, i was taking a look at the G933/935 implementation it looks like the correct size is 20 bytes for the `HIDPP_LONG_MESSAGE` of the battery request, but the send_sidetone message has no 0x11 [`HIDPP_LONG_MESSAGE`] flag, i suppose it will be fine with a message lenght of `HIDPP_LONG_MESSAGE_LENGTH` since before the MSG_SIZE edit it was 12 bytes long, but i kept it separated.
It should be probably fine to use `HIDPP_LONG_MESSAGE_LENGTH` for both the battery request and sidetone

Sapd:
I find it kind of weird, that one feature uses HIDPP_LONG_MESSAGE, but the other one 64 bytes, which isn't defined in HID++.
However I find it unlikely that the Sidetone is HIDPP_LONG_MESSAGE, because it must begin with HIDPP_LONG_MESSAGE / 0x11 then (first byte is the reportid).

It will most likely just make an problem on Windows, so I will merge it.

PR #72, opened 2020-03-10
udev: give rights only to seated user

xdbob:
The `uaccess` udev TAG allows to dynamically set permission to
physically connected users thus we can avoid giving RAW access to the
devices to everyone.

Note: I also merged the rules into a single file for simplicity and moved it from `50-*` to `70-*` because it is the norm used in multiple projects settings hidraw permissions.


Sapd:
Thanks for cleaning up the udev rules.

(The Failed build is some not related problem with os x)

PR #75, opened 2020-04-01
Added Corsair void pro surround

Doomsdayrs:
`-n 0` & `-n 1` working

`-s` is not working. Error:
```
ioctl (SFEATURE): Broken pipe.
Failed to set sidetone. Error: -1: (null)
```
`-l 0|1` does nothing

`-b` provides a state of `CHARGING`


Sapd:
Theoretically everything should work, either something doesn't fit in the implementation yet, something is locally not right - or the headset has a different protocol.

PR #76, opened 2020-04-10
Add Corsair VOID RGB ELITE WIRELESS

jb-leger:
This PR add the the _VOID RGB ELITE WIRELESS_ headset.

This works without problems for me.

_Please note:_ I format the `PRODUCT_IDS` array with one value perl line with a trailing comma, in order to add only a line when a device is added in the future and avoid potential conflicts. If you prefer I can keep the original formatting.


Sapd:
> _Please note:_ I format the `PRODUCT_IDS` array with one value perl line with a trailing comma, in order to add only a line when a device is added in the future and avoid potential conflicts. If you prefer I can keep the original formatting.

Yes this looks much cleaner. 

I merged it

PR #77, opened 2020-04-21
Add battery status for G930

tobiaspc:
I simply ported (copied actually) it from https://github.com/Roliga/g930-battery-percentage

Seems to work quite good. 

Issues:
- Charging status is not read from the device
- Runtime / Charge time is not read
- During Charge, battery goes over 100%

I believe the code directly reads the voltage, maybe someone can improve that in a later step.

This is my first pullrequest, is everything in order?

### DIFF
--- a/src/devices/logitech_g930.c
+++ b/src/devices/logitech_g930.c
@@ -1,4 +1,5 @@
 #include "../device.h"
+#include "../utility.h"
 #include "logitech.h"
 
 #include <stdio.h>
@@ -7,8 +8,6 @@
 
 #include <hidapi.h>
 #include <string.h>
-
-#include <errno.h>
 
 #define BATTERY_MAX 91
 #define BATTERY_MIN 44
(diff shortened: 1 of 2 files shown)


src/devices/logitech_g930.c
```diff
@@ -18,19 +31,57 @@ void g930_init(struct device** device)
...
+        fprintf(stderr, "Failed to send feature report: %ls\n", hid_error(device_handle));
+        return -1;
+    }
+
+    for (i = 0; i < 3; i++) {
+        res = hid_get_feature_report(device_handle, buf, MSG_SIZE);
+
+        if (res < 0) {
+            fprintf(stderr, "Failed to get feature report %d: %ls\n", i, hid_error(device_handle));
+            return -1;
```
Sapd, line 77:
```suggestion
            return res;
```

The main function does the error handling and will also print the error.

src/devices/logitech_g930.c
```diff
@@ -18,19 +31,57 @@ void g930_init(struct device** device)
...
+        }
+
+        if (i < 2) {
+            usleep(100000);
+        }
+    }
+    unsigned int batteryPercent = (buf[13] - BATTERY_MIN) * 100 / (BATTERY_MAX - BATTERY_MIN);
+
+    return (int)(batteryPercent);
```
Sapd, line 86: You can use map (include "../utility.h" for that)

```suggestion
    int batteryPercent = map((int)buf[13], BATTERY_MIN, BATTERY_MAX, 0, 100);
    
    return batteryPercent;
```

Sapd:
Thanks for your submission, I made some comments.

I think the battery goes over 100%, because the battery of the owner of the other repository was already quite old and wasn't able to charge it completely.
You can probably set BATTERY_MAX safely to 100.

tobiaspc:
Thanks for the quick review, I think everything should now be correct. 
My device is also quite old (from ~2012). 
After a full day of charging, the program reports 100% with the current upper limit of 91. So either that limit is correct, or my battery is also dying. 
Since I am about to replace the battery in a few days, I can verify the upper limit then with a fresh one.

If you want, you can pull this branch now. Either way, i am willing to investigate the protocol further, in order to get charging state / remaining time.

Sapd:
Yes, I included the current state.

PR #81, opened 2020-04-28
Add capabilitiy for setting inactive time

centic9:
Add a new capability CAP_INACTIVE_TIME to set time until a device turns itself off. 
Implement this for SteelSeries Arctis which supports this.

### DIFF
--- a/src/devices/steelseries_arctis.c
+++ b/src/devices/steelseries_arctis.c
@@ -100,25 +100,13 @@ static int arctis_request_battery(hid_device* device_handle)
 
 static int arctis_send_inactive_time(hid_device* device_handle, uint8_t num)
 {
-    int ret = -1;
-
     // as the value is in minutes, mapping to a different range does not make too much sense here
     // the range of the Arctis 7 seems to be from 0 to 0x5A (90)
     // num = map(num, 0, 128, 0x00, 0x5A);
 
-    unsigned char* buf = calloc(31, 1);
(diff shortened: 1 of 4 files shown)


src/devices/steelseries_arctis.c
```diff
@@ -89,3 +91,28 @@ static int arctis_request_battery(hid_device* device_handle)
...
+
+    const unsigned char data[3] = { 0x06, 0x51, num };
+
+    memmove(buf, data, sizeof(data));
+
+    ret = hid_write(device_handle, buf, 31);
+
+    free(buf);
+
+    return ret;
```
Sapd, line 117: I'd suggest doing it this way (I also plan to do it on further code changes).

Note that when initialising on stack via initialiser list with the curly brackets in c or c++, all missing values are automatically set to 0.
```suggestion
    // as the value is in minutes, mapping to a different range does not make too much sense here
    // the range of the Arctis 7 seems to be from 0 to 0x5A (90)
    // num = map(num, 0, 128, 0x00, 0x5A);

    uint8_t data[31] = { 0x06, 0x51, num };
    return hid_write(device_handle, buf, 31);
```

centic9: I pushed changes accordingly and also another change as there is a "save_state" command which seems to be necessary for SteelSeries devices to actually remember the changed setting across power on/off.

PR #84, opened 2020-05-01
Adjust introductory section and add invactive time

centic9:
The tool is not only about sidetone any more, so add a more general introduction

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,9 @@
 ## Summary
 
+A tool to control certain aspects of USB-connected headsets on Linux. Currently support is provided for adjusting sidetone, getting battery state, controlling LEDs and setting the inactive time. See below for which headset supports which feature.
+
+### Sidetone
+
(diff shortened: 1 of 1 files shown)


Sapd:
Yeah the text was a bit dated.
Grammerly says, that some commas are missing in your suggestion:

> A tool to control certain aspects of USB-connected headsets on Linux. Currently, support is provided for adjusting sidetone, getting battery state, controlling LEDs, and setting the inactive time. See below for which headset supports which feature.

centic9:
Sorry, I don't use that service and do not want to install it. Also I am not a native speaker, so not sure which ones exactly you are missing here? :)

Sapd:
It's an online service with AI. But I wasn't sure bc I'm not native speaker; I quoted the corrected version above, here again:

> A tool to control certain aspects of USB-connected headsets on Linux. Currently, support is provided for adjusting sidetone, getting battery state, controlling LEDs, and setting the inactive time. See below for which headset supports which feature.

(But it's not really important, I just waited for your opinion)

centic9:
Ah, ok, didn't notice that, I updated the branch now.

PR #90, opened 2020-05-17
Add support for Logitech G PRO

jomority:
Sidetone is working fine for me. The headset does not support any of the other features.

The Logitech G PRO X is very similar to the G PRO and I included its product id as a comment. But since I can't test it, I don't want to include it. If someone gets his hands on the device it might be very simple to add it as well.

### DIFF
--- a/src/devices/logitech_gpro.c
+++ b/src/devices/logitech_gpro.c
@@ -1,0 +1,36 @@
+#include "../device.h"
+#include "../utility.h"
+#include "logitech.h"
+
+#include <string.h>
+
+static struct device device_gpro;
+
+static const uint16_t PRODUCT_ID = 0x0aa7; // PRO X: 0x0aaa
+
+static int gpro_send_sidetone(hid_device* device_handle, uint8_t num);
+
+void gpro_init(struct device** device)
+{
+    device_gpro.idVendor = VENDOR_LOGITECH;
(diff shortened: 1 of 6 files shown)


src/devices/logitech_gpro.c
```diff
@@ -0,0 +1,36 @@
...
+    device_gpro.send_sidetone = &gpro_send_sidetone;
+
+    *device = &device_gpro;
+}
+
+static int gpro_send_sidetone(hid_device* device_handle, uint8_t num)
+{
+    // num = map(num, 0, 128, 0, 100);
+    if (num > 0x64)
+        num = 0x64;
```
jomority, line 31: I'm not completely sure which solution is better.
The wiki example uses the `map` function, but the implementation in `logitech_g633_g933_935.c` uses this capping approach.

Sapd: Map is to be preferred. Because the user may supply a value from 0 - 128, and the cap simply ignores values after 100, which is confusing for the user.

jomority: Ok, I will change it to map.

Sapd:
Note that the CI will fail, because the #includes in device_registry.c are not sorted (only something minor - but I try to write in accordingly to clang tidy style rules).

jomority:
Yes, I just noticed and fixed it. If the checks complete successfully, I think it is ready for merge.

Issue #65, opened 2020-01-12
G633 side lights doesn't turn off


Any0ne22:
When trying to turn the lights off on my G633 headset, only the light strips on the back are turning off. The G logo is still enlightened.

KrDimitrov:
Same on G933, will  look into it

Sapd:
Fixed by #91

PR #91, opened 2020-05-21
G933 Lights fix

KrDimitrov:
The light setting now controls both the strips on the back and the side logos. 

**Resolves** #65

**Needs Further Testing**

Sometimes my G933 ignored one of the requests. A delay between the requests seems to fix that. I've also only tested this on windows, maybe not even an issue on Linux?

Also needs testing for G633 and G935. Maybe is uses different bytes? I don't have a G633/G935 and cant test...


Sapd:
> Sometimes my G933 ignored one of the requests. A delay between the requests seems to fix that. I've also only tested this on windows, maybe not even an issue on Linux?

Also happened for me on another headset on Linux. Seems that there must be some small time between two requests.

Sapd:
I think I can merge it

PR #94, opened 2020-05-31
Add capability to read the value of the "chat-mix" slider on SteelSeries Arctis

centic9:
This is a feature of the SteelSeries Arctis headset when using the bundled Windows software which allows to adjust how Game-Sound is mixed with live audio-chat via a slider on the Headset.

Reading this value allows arbitrary integrations to be built, i.e. you could use this to control play-speed in your audio player or any other integration that can be done via reading the setting of a hardware slider.

Changes:
* Added a new capability "CAP_CHATMIX_STATUS"
* Added option "-m" to read it
* Implemented the capability for the Steelseries Arctis
* Additionally fixed a few typos in the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ talking. This differs from a simple loopback via PulseAudio as you won't have an
 
 ## Supported Headsets
 
-- Corsair Void (Every version*, regardless wether Pro or Wired)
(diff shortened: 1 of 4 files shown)


Sapd:
Thanks for your contribution

PR #97, opened 2020-06-09
Add support for Corsair HS70 Pro.

dries007:
Since all of the controls seem to work (but no lights, since there are none) I didn't want to create a new file. Even though this model isn't from the "Void" range.

Edit: Thanks for this work btw, I'm so happy I finally have a nice Linux tool.


Sapd:
Thank you, somehow I missed that PR

PR #98, opened 2020-07-22
Arctis 1 support

matthewbaggett:
Partially working Arctis 1 support

Battery level doesn't work. Not sure why yet. But atleast I can set sidetalking from linux now.

Forked steelseries_arctis.c into two, since the arctis 1 seems a bit too different.

### DIFF
--- a/src/device_registry.c
+++ b/src/device_registry.c
@@ -6,11 +6,12 @@
 #include "devices/logitech_g633_g933_935.h"
 #include "devices/logitech_g930.h"
 #include "devices/logitech_gpro.h"
-#include "devices/steelseries_arctis.h"
+#include "devices/steelseries_arctis_1.h"
+#include "devices/steelseries_arctis_7.h"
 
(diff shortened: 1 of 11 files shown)


Sapd:
> Forked steelseries_arctis.c into two, since the arctis 1 seems a bit too different.

Makes definitely sense.

> Battery level doesn't work. Not sure why yet. But atleast I can set sidetalking from linux now.

Maybe we should disable it then for the Arctis 1 for now.

Thank you for contributing 👍

denis-exo:
The arctis_1_request_battery must be commented out, or CI will fail.
(I don't have permissions to push to your PR)

matthewbaggett:
Yess! :+1: I did a contribute.

hadess:
The battery checking should be working, if this alternative implementation also does:
https://github.com/atagulalan/arctis-battery-percentage/blob/master/percentage.js#L32-L36

Issue #56, opened 2019-12-10
Logitech g933-935 Battery issue!


ziadkiwan:
Hello,
i have ubuntu 19.10 when i use HeadsetControl -b i get the following:

Found Logitech g933-935 Wireless!

Failed to request battery. Error: -24: (null)

Sapd:
Which is your headset? G933 or G935

ziadkiwan:
G933, if you want i can give you anydesk access, so you check ?

Sapd:
I can't really help with Anydesk, as it could be an implementation issue or some other things.

Does Sidetone work?

ziadkiwan:
Hello,

I complied the source, and everything was successful, all the controls fails for some strange reason, if you have any suggestions i would be glad to test or implement!

Sapd:
I just looked it up, error 24 is `EMFILE: Too many open files`
I'm not really sure how this can happen. 
You can try increasing the file limit beforehand: `sudo sysctl -w fs.file-max=100000` does this temporarily.

Alternatively, if this does not work, try opening `cmake_modules/Findhidapi.cmake` and in line 27 replace the line by `NAMES hidapi hidapi-libusb)`

I guess there is some kind of bug in a library in your system.

ziadkiwan:
Hello, 

You are right! when i did `sudo sysctl -w fs.file-max=100000` things worked, well as you know Linux deals with everything as files, and there is a limit for number of opened files, but I'm 100% sure that i did not open this amount, maybe the library you are relaying on does not have compatibility with Ubuntu 19.10 ?
I will remove the build folder and try editing the `cmake_modules/Findhidapi.cmake` maybe it will be a better permanent fix!,

i will get back to you once it's done!

Thanks for the help!

Sapd:
What surprised me is that error 24 is process specified (EMFILE), as error 23 is global (ENFILE) as far as I understand. So the process itself opened up too many.

So either your file limit is ridiculously low (check with `sysctl fs.file-max` after a reboot). Or the hidapi library your distribution ships with, has a bug.

Instead of editing the cmake file (this will change the library to the libusb implementation of hidapi), you can also write the sysctl parameter permanently into /etc/sysctl.conf 
However it would be interesting anyway, how changing the library would affect the bug.

ziadkiwan:
Hello Guys, i found the bug that was caused the issue, if you call HeadsetControl -b while it's charging you get this error Failed to request battery. Error: -24: (null).

[...]

PR #111, opened 2020-11-03
Update Findhidapi.cmake

fr33-man:
This should fix cmake errors and make headsetcontrol -b work. #56

### DIFF
--- a/cmake_modules/Findhidapi.cmake
+++ b/cmake_modules/Findhidapi.cmake
@@ -37,7 +37,7 @@ find_path(HIDAPI_INCLUDE_DIR
 	hidapi)
 
 include(FindPackageHandleStandardArgs)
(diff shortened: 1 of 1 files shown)


Sapd:
Not really sure why this makes a difference, however CI checked all platforms. And looking into the cmake file, it is also referenced as lower case anyway, so should be good to merge

fr33-man:
When it's typed with uppercase letters it prints this:
`CMake Warning (dev) at /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:273 (hidapi):
  The package name passed to `find_package_handle_standard_args` (HIDAPI)
  does not match the name of the calling package (hidapi).  This can lead to
  problems in calling code that expects `find_package` result variables
  (e.g., `_FOUND`) to follow a certain pattern.`

Issue #112, opened 2020-11-13
Corsair Void Pro Wireless error when low on battery


elril0000:
Whenever the battery is lower than 20% the program outputs the following:

Found Corsair Headset Device!

Failed to request battery. Error: -100: Success

After some debugging I found that the data_read[4] value reports 2.
When adding 2 to the condition the program runs fine

Sapd:
Fixed by #114

PR #113, opened 2020-11-13
Fix for Corsair Void Pro Wireless low battery error

elril0000:
fix for <a href='https://github.com/Sapd/HeadsetControl/issues/112'>issue #112</a>

### DIFF
--- a/src/devices/corsair_void.c
+++ b/src/devices/corsair_void.c
@@ -115,7 +115,11 @@ static int void_request_battery(hid_device* device_handle)
         return BATTERY_CHARGING;
     }
 
+<<<<<<< HEAD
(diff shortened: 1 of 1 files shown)


src/devices/corsair_void.c
```diff
@@ -115,7 +115,7 @@ static int void_request_battery(hid_device* device_handle)
         return BATTERY_CHARGING;
     }
 
-    if (data_read[4] == 1) {
+    if (data_read[4] == 1 || data_read[2]) {
```
Sapd, line 118: But then this check must be
`if (data_read[2])`

because data_read[4] == 1 is redundant

elril0000: I made a typo I meant to write data_read[4] == 2

PR #118, opened 2020-12-04
Add support for the SteelSeries Arctis 1 Wireless for XBox

hadess:
Battery level _seems_ to be correct. I couldn't/didn't test the sidetone level (I have no idea what that is), and the timeout didn't work either, but I wonder how it detects usage (I had the mic disabled on the headset, but PulseAudio running).

### DIFF
--- a/src/device.h
+++ b/src/device.h
@@ -51,7 +51,7 @@ struct device {
     uint16_t idUsage;
 
     /// Name of device, used as information for the user
-    char device_name[32];
+    char device_name[64];
 
     /// Bitmask of currently supported features the software can currently handle
     int capabilities;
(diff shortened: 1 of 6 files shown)


Sapd:
>  I couldn't/didn't test the sidetone level (I have no idea what that is)

When Sidetone is one and you speak into the microphone, then you should hear yourself (regardless wether a application is listening or not). Is this the case? If not the headset may not support is and you can remove the device_arctis.capabilities flag for sidetone and the respective function.

hadess:
> When Sidetone is one and you speak into the microphone, then you should hear yourself (regardless wether a application is listening or not). Is this the case? If not the headset may not support is and you can remove the device_arctis.capabilities flag for sidetone and the respective function.

Oh, putting it at 128 without speaking amps up the white noise, and switching between say, 100 and 16, changes the tone of my voice. So I say this looks like it's working.

PR #129, opened 2021-01-11
adds sidetone support for Logitech G Pro X Wireless

thoraxe:


Sapd:
I noticed that the packet is the same like in logitech_gpro.c

you can just add the ID of your headset to the existing implementation (see into the .c file of Corsair void for an example) in logitech_gpro.c

thoraxe:
@Sapd 10-4, I will work on that.

Do you have any interest in me trying to figure out other things like battery level, surround sound settings, etc?

Sapd:
Feel free to ask any questions, but make sure to take a look here, there is a full tutorial: https://github.com/Sapd/HeadsetControl/wiki/Development

thoraxe:
Updated PR by combining Pro and Pro X into a single file as requested, removed prox files.

I'll ask the questions about battery, etc. via an issue.

PR #136, opened 2021-02-17
Keep recognition separated and device handling as one

angeldeejay:
Fix devices handling to keep requests as one for all matching devices:

- G633
- G635
- G733
- G933
- G935

But keeping recognition separately

### DIFF
--- a/src/device_registry.c
+++ b/src/device_registry.c
@@ -3,8 +3,12 @@
 #include "devices/corsair_void.h"
 #include "devices/logitech_g430.h"
 #include "devices/logitech_g533.h"
-#include "devices/logitech_g633_g933_935.h"
+#include "devices/logitech_g633.h"
+#include "devices/logitech_g635.h"
+#include "devices/logitech_g733.h"
 #include "devices/logitech_g930.h"
+#include "devices/logitech_g933.h"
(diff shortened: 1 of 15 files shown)


angeldeejay:
image

Possibly-Matt:
Works great for me! Accurately found my G733

Sapd:
Im quite skeptical about the benefit this would provide. Instead we could simply use one file, then insert there all ids.
 
Separating it like this (actually one could say, you are using the same pattern as C++ parent-child relationships), would make sense, when one device has some differences but also at the same time some common code. In this case however, as I see it, all the code is same.

> But keeping recognition separately

Actually we basically recognise or match the _implementation_ instead of the real device. 

We could theoretically give the implementation a name (just like now e.g. logitech g9xx) AND display the device name by just printing the name HIDAPI returns.



Also note, that by accepting this, we would also have to do the same thing with corsair, to keep code consistent. That means actually 14 new files, which basically don't do anything - instead having a different string. 

To print the correct name of the devices, we can modify the string here https://github.com/Sapd/HeadsetControl/blob/5fecb002586b72e07b35912a74cd2212310c38b0/src/main.c#L59 and include hid_device_info.manufacturer_string and hid_device_info.product_string

PR #140, opened 2021-03-09
Add sidetone support for Logitech Zone Wired headset

s3lph:
This PR adds support for the Logitech Zone Wired headset and configuring its sidetone volume.


simonbaird:
See also #174 .

PR #143, opened 2021-03-25
Allow to query for capabilities of current headset

centic9:
When invoking headsetcontrol in other tools it would be useful to be able to dynamically adjust to the capabilities of the currently detected headset.

Therefore this PR adds a commandline option "-?" which prints out a list of things that will work for the current headset

```
$./headsetcontrol -?
Found SteelSeries Arctis (7/Pro)!

Supported capabilities:

* set sidetone
* read battery level
* switch lights
* set inactive time
* read chat-mix
```

Short output is also supported to make it easy to use this in other tools:

```
$./headsetcontrol -? -c
sblim
```


Sapd:
Good work

Issue #127, opened 2021-01-10
Support for Roccat Elo 7.1 Air?


Sebiann:
If it's possible, could it be added?

iss0:
I am interested in the Roccat Elo 7.1 Air as well. When I find the time I will have a look with Wireshark and the Roccat Windows software.

lukasgr90:
I've started working on this. So far I'm able to switch the LEDs between off and white.

lukasgr90:
I've also reverse engineered and implemented setting the automatic standby time. But it does only seem to apply after restarting the headset. I need to further test it.

I won't implement the sidetone, since controlling this is mainly not done over the USB HID protocol. Instead the official software controls the sidetone basically by sending USB control commands (layer under HID) to the audio interfaces. The latter are used by the Linux audio drivers, which luckily already support controlling the sidetone! Thus a Linux user can do following with amixer:
* First find out the `ALSA-CARD-NUMBER`, e.g. with `pactl list` and looking for _alsa.card_ (better options are welcome :-) )
* Then to switch it on/off: `amixer -c <ALSA-CARD-NUMBER> cset numid=5,iface=MIXER,name='Main Configuration Playback Switch' <ON-OFF>` with `<ON-OFF>` being one of "on" or "off"
* To set the level: `amixer -c <ALSA-CARD-NUMBER> cset numid=6,iface=MIXER,name='Main Configuration Playback Volume' <VALUE>` with `<VALUE>` being a value between *0* (off) and *5632* (highest effect)

I partially understood how the battery levels are communicated. 
- After a restart the Headset sends the command: 0xE6 0x06 0x03 0x00 **0x04** [... zeros, except 24th Byte is 37]
- The official software sets the battery indicator to 100%
- Then the Headset sends the command: 0xE6 0x06 0x03 0x00 **0x03** [... zeros, except 24th Byte is 37]
- The official software sets the battery indicator to 75%
- Then Headset sends the command: 0xE6 0x06 0x03 0x00 **0x02** [... zeros, except 24th Byte is 37]
- The official software sets the battery indicator to 50%
... probably this "play" goes further, if my battery is used more. If we don't find a possibility to poll the battery state, we would need a daemon process. I'm not sure, how to fit this with HeadsetControl's design :-/.

Sapd:
[...]

PR #147, opened 2021-04-23
Added particular support for the Roccat Elo 7.1 Air.

lukasgr90:
These changes add the new device "Roccat Elo 7.1 Air". For now only control of the LEDs is supported. More might come.

Partially implements #127


src/devices/roccat_elo_7_1_air.c
```diff
@@ -0,0 +1,97 @@
...
+ */
+static int send_receive(hid_device* hid_device, const unsigned char* data, 
+			int size, unsigned char* out_buffer) 
+{
+    struct timespec ts;	
+    unsigned char send_buffer[64];
+    int r;
+
+    memset(send_buffer, 0, sizeof(send_buffer));
+    memcpy(send_buffer, data, size);
```
Sapd, line 49: Thank you!

Just a question regarding your send_receive function. As I dont quite understand the sense.

E.g. when you do this:
```
     unsigned char cmd92[64] = {0xff, 0x02};

     hid_write(hid_device, cmd92, sizeof(cmd92)/sizeof(cmd92[0]));
```

The first line, will automatically pad everything with 0s. Thats defined in the C-Standard. So you dont actually have to clear the array with memset beforehand.

lukasgr90: Thank you for the feedback!

If that's defined in the C-Standard, I'll of course remove this as obsolete.

It is quite a time ago that I actively programmed C / Assembler and just remembered, that I should rather be defensive about such things. ;-)

src/devices/roccat_elo_7_1_air.c
```diff
@@ -0,0 +1,110 @@
...
+static const uint16_t PRODUCT_ID = 0x3a37;
+
+static int elo71Air_switch_lights(hid_device* hid_device, uint8_t on);
+static int elo71Air_send_inactive_time(hid_device* hid_device, uint8_t num);
+
+void elo71Air_init(struct device** device)
+{
+    device_elo71Air.idVendor = VENDOR_ROCCAT;
+    device_elo71Air.idProductsSupported = &PRODUCT_ID;
+    device_elo71Air.numIdProducts = 2;
```
lukasgr90, line 19: After looking at usage of this parameter at other device's code I think that I did not set it correctly!?

@Sapd Can you please have a look and give a hint, how it is used properly (if necessary)?

src/devices/roccat_elo_7_1_air.c
lukasgr90: I hope that I corrected the code properly.

lukasgr90 (commented):
Thanks to @iss0 for contributing the udev rules!

src/devices/roccat_elo_7_1_air.c
Sapd: Yep, it must be 1.
It simply says, how big the array device_elo71Air.idProductsSupported is. As you only provide one id, its size is 1.
When you specify 2, the loop in device_manager.c would run into some other memory and possibly crash or do some other unexpected things.

PR #161, opened 2021-06-10
Add support for Logitech G433 (Sidetone)

Athemis:
This pull requests adds support for the Logitech G433 headset (sidetone). This is basically the same thing as G432, just copied the files and swapped out names and USB ID. Tested only on Linux.


Athemis:
Ok, probably bad phrasing on my side: It's not some kind of "upgraded" G432, it's a completely different look and feel. However since sidetone setting seems to work identical I'll just merge those two ;)

Sapd:
Yeah there are often headsets with same protocol.

The sense of the source files is practically not to group headsets, but to group protocols and code. Otherwise we would have double code, and additional effort and error sources

PR #174, opened 2021-09-29
Add support for Logitech Zone 750

simonbaird:
I think it's identical or very nearly identical to the Zone Wired,
so this patch just adds another product id to the existing device.

I tested it with my headset and I can confirm the sidetone level,
the voice prompt toggle, and the rotate to mute toggle, all work for
me as expected.

(The part number for the Zone 750 according to logitech.com is
981-001103. There are two part numbers for the Zone Wired, "UC Zone
Wired", 981-000876, and "MSFT Teams Zone Wired", 981-000871.)


simonbaird:
https://www.logitech.com/en-us/products/headsets/zone-750-wired-usb.981-001103.html FYI.

Compare it to https://www.logitech.com/en-us/products/headsets/zone-wired-noise-canceling.html .

simonbaird:
I pushed a small revision to make the linter happy.

simonbaird:
Rebased on origin/master, no other change.

Issue #194, opened 2021-11-18
Logitech Zone Wired setting sidetone disables it


culebrasolida:
### Description

Whenever I try to set sidetone volume for a Logitech Zone Wired headset with:
`sudo headsetcontrol -s --50
Found Logitech Zone Wired/Zone 750!
Success!
`

That completely disables sidetone and headset has to be plugged back in to get sidetone back.

### Headset Name

Logitech Zone Wired

### On which OS does the problem happen?

Linux

### Device information

<details>
 <summary>Detailed Device Information</summary>

 ```
Device Found
  VendorID: 0x046d
 ProductID: 0x0aad
  path: /dev/hidraw3
  serial_number: 2017ALQ00SV2
  Manufacturer: Logitech
  Product:      Zone Wired
  Interface:    3
  Usage-Page: 0xc Usageid: 0x1

Device Found
  VendorID: 0x046d
 ProductID: 0x0aad
  path: /dev/hidraw3
  serial_number: 2017ALQ00SV2
  Manufacturer: Logitech
  Product:      Zone Wired
  Interface:    3
  Usage-Page: 0xff99 Usageid: 0x4

Device Found
  VendorID: 0x046d
 ProductID: 0x0aad
  path: /dev/hidraw3
  serial_number: 2017ALQ00SV2
  Manufacturer: Logitech
  Product:      Zone Wired
  Interface:    3
  Usage-Page: 0xb Usageid: 0x5

Device Found
  VendorID: 0x046d
 ProductID: 0x0aad
  path: /dev/hidraw3
  serial_number: 2017ALQ00SV2
  Manufacturer: Logitech
  Product:      Zone Wired
  Interface:    3
  Usage-Page: 0xb Usageid: 0x6

Device Found
  VendorID: 0x046d
 ProductID: 0x0aad
  path: /dev/hidraw3
  serial_number: 2017ALQ00SV2
  Manufacturer: Logitech
  Product:      Zone Wired
  Interface:    3
  Usage-Page: 0xff00 Usageid: 0x5

Device Found
  VendorID: 0x046d
 ProductID: 0x0aad
  path: /dev/hidraw3
  serial_number: 2017ALQ00SV2
  Manufacturer: Logitech
  Product:      Zone Wired
  Interface:    3
  Usage-Page: 0xff30 Usageid: 0x6

Device Found
  VendorID: 0x046d
 ProductID: 0x0aad
  path: /dev/hidraw3
[...]

PR #195, opened 2021-12-02
Added timeout parameter for reading, default is 5s

### LINKED ISSUE #193: Battery status freeze
### Description

When using headsetcontrol on a regular basis (e.g. in scripts) it tends to freeze after the device name output or directly when using the short output after a some executions. This was already reported in #125, but doesn't seem to be limited to a specific device, in my case it's a SteelSeries Arctis 9.  Instead of

>Found SteelSeries Arctis 9!
>Battery: 31%
>Success!

I just get a

>Found SteelSeries Arctis 9!

and no further output. Sometimes this happens after the second or third execution of headsetcontrol, sometimes it works for 100 executions without a problem.

This makes projects like the mentioned [headsetcontrol-notificationd](https://github.com/Manawyrm/headsetcontrol-notificationd) unusable, as the script locks up in the background.

If one instance of headsetcontrol hangs after requesting the battery status, all other instances querying the battery status will freeze as well. Terminating all instances and waiting for a bit seems to solve the problem and a new instance of headsetcontrol will display the battery status again -- until a future request will freeze again.

System information:

> Debian stable
> Linux 5.10.70
> udev 247.3
> libhidapi 0.10.1

### On which OS does the problem happen?

Linux

Sapd:
Also:
- Improve parameters with numbers, check if they are really numbers
- Pass error codes to program return
- On timeout, program returns -101

Fixed #193

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ talking. This differs from a simple loopback via PulseAudio as you won't have an
 
 ## Supported Headsets
 
-- Corsair Void (Every version*, regardless whether Elite, Pro, HS70, Wireless or Wired)
+- Corsair **Void** (Every void-version*, regardless whether Elite, Pro, HS70 Wireless)
   - Sidetone, Battery (for Wireless), LED on/off, Notification Sound
 - Logitech G430
   - No support in current version (Last working on macOS in commit 41be99379f)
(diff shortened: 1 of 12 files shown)


Sapd:
Also fixes number parsing / errors, probably leading to errors like #194

PR #196, opened 2021-12-30
Arctis 1 Wireless - fixed inactive time functionality and added Windows Support

alisadah:
Not sure if the addresses are different for the Xbox variant, but my Arctis 1 Wireless (firmware 1.6.0.0) uses 0x53 instead of 0x51 for setting the inactivity shutdown timer. 

Also added the correct usage page/ID for Windows support and it's completely functional on Windows now

### DIFF
--- a/src/devices/steelseries_arctis_1.c
+++ b/src/devices/steelseries_arctis_1.c
@@ -29,9 +29,9 @@ void arctis_1_init(struct device** device)
     strncpy(device_arctis.device_name, "SteelSeries Arctis (1) Wireless", sizeof(device_arctis.device_name));
 
     device_arctis.capabilities                           = B(CAP_SIDETONE) | B(CAP_BATTERY_STATUS) | B(CAP_INACTIVE_TIME);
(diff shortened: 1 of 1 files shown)


src/devices/steelseries_arctis_1.c
```diff
@@ -112,7 +112,7 @@ static int arctis_1_send_inactive_time(hid_device* device_handle, uint8_t num)
     // the range of the Arctis 7 seems to be from 0 to 0x5A (90)
     // num = map(num, 0, 128, 0x00, 0x5A);
 
-    uint8_t data[31] = { 0x06, 0x51, num };
+    uint8_t data[31] = { 0x06, 0x53, num };
```
Sapd, line 115: Im not 100% sure about this change. We may need someone else with another headset to test if it still works for them. In doubt we need to put it inside a different device file.

joghert: I have an Arctis 1 Wireless, and can confirm these settings are correct. The inactivity shutdown timer is working now

`Bus 003 Device 007: ID 1038:12b3 SteelSeries ApS SteelSeries Arctis 1 Wireless`

Sapd: Maybe they were not correct from the beginning

Issue #131, opened 2021-01-12
Pro X / Pro battery support


thoraxe:
This is an issue to discuss figuring out battery support on the Pro / Pro X wireless headsets.

The G933-esque battery support came from https://github.com/ashkitten/g933-utils

That repo includes a Wireshark Lua plugin to help analyze the Logitech HID+ protocol, but it seems to have a few shortcomings with the newer G Hub software and the Pro X.

https://github.com/ashkitten/g933-utils/issues/17

I didn't see anywhere in the "known" messages where the battery status was queried by the G-Hub software. I tried the following:

* start wireshark with dongle connected but headset power off
* open g-hub application
* turn on headset

At this point, the headset shows up as connected in G-Hub and the battery level appears. So, somehow the app is determining the battery percentage somehow. 

The G933 implementation has a get voltage function that it then converts to a percentage:
https://github.com/Sapd/HeadsetControl/blob/master/src/devices/logitech_g633_g933_935.c#L51-L89

However, as I mentioned earlier, the Pro X doesn't appear to ever be sent any kind of get battery request that's "known" at this point.

In the issue I filed with ashkitten's repo, you can see the Lua issues I'm having.

If you have any suggestions or ideas, let me know. I have the device so it's easy for me to poke at it.

Sapd:
I'd simply try it without that plugin. Wireshark is quite powerful and has a lot of filters.

thoraxe:
Right, but the plugin is what is dissecting the packet capture and explaining what is being sent to/from the headset.

Here's a capture of turning the headset on and waiting and then turning it off.
https://drive.google.com/file/d/11r9voCayKynHl0V2QVMm9RHaIVcexw_p/view?usp=sharing

markpash:
@thoraxe I just downloaded your pcap and went through it as I don't feel like booting into windows. I think I may have found the message that requests the battery info.

Luckily the response is identical to the G933/G935 response, so after a few attempts, I think I found the correct request and tested it with my headset. It's able to correctly identify if it's charging or not. I'm not sure if the voltage calculation is correct since I just copied it from the G933 code. I can make a PR for this, but this may need to be a separate device, as the normal wired headset may not support getting battery info.

@Sapd would you be willing to review a PR if I make one?

Sapd:
> @Sapd would you be willing to review a PR if I make one?

Sure, feel free to open up a PR any time. Things can then still be completely changed or improved.

PR #197, opened 2022-01-05
PRO X Wireless Headset

jaypikay:
Logitech devices seem a bit strange when reporting their voltage. As far as I could see the PRO X reports the voltage capacity.

- Battery status
  - charging
  - discharging (returns voltage)
- Set inactivity time


src/devices/logitech_gpro.c
```diff
@@ -2,8 +2,14 @@
 #include "../utility.h"
 #include "logitech.h"
 
+#include <unistd.h>
+
+#include <math.h>
 #include <string.h>
 
+#define MSG_SIZE    20
```
Sapd, line 10: The MSG_SIZE is not used anywhere, so we can probably remove it

jaypikay:
Just noticed that #131 is related.

jaypikay:
Just noticed some quirks with the voltage. When it gets lower the percentage isn't correct anymore

Sapd:
Yes, you used a linear formular for the voltage. However LI-Ions usually hold their voltage for a long time (slowley decreasing) and then suddenly decreasing when they are nearing under 20%.

E.g. here are examples: https://hobbygraderc.com/lipo-voltage-chart/

You can use a polynomial regression calculator to make a mathematical curve: https://arachnoid.com/polysolve/ (the website will even generate C code)

Or you can use simple splines:
Simply set p and v to your liking, which comes nearest to your headset
```
static float estimate_battery_level(uint16_t voltage) {
    int p[] = { 100, 60, 10, 0 };
    int v[] = { 4200, 3870, 3690, 3270 };
    int size = 4;
    
    float percent = 0;
    for(int i = 0; i < size; ++i) {
        // if >= then 100%
        if(voltage >= v[i]) {
            percent = p[i];
            break;
        } 
        
        // if not last
        if(i < size-1 && voltage >= v[i+1]) {
            percent = p[i+1] + (voltage-v[i+1])/(v[i]-v[i+1])*(p[i]-p[i+1]);
            break;
        }
       return percent;
}
 ```

We can put the function in utility.h with p and v as parameters, so every headset can use it if it wants to use that method.

jaypikay:
Hey thank you, for you post.

1. Inactivity timeout: I don't know, haven't looked into this, but it's an idea. I can try to look into it.
2. Need to look into it.
3. Does not look like this could be a problem, the head set manages the standby by itself.
4. There shouldn't be any differences. Except the microphone stuff. Logitech uses this *Blue* thing stuff, but it should be possible to achieve similar microphone quality using different audio filter tools, for Pulseaudio/Pipewire or Alsa and Jack. I do not have more details on this topic, as I do not care.
5. The "forks" are metal

Bananaman:
@jaypikay Thank you so much for your answers. I apologize, I thought I had replied earlier. I am really grateful that you took the time.

I'm glad to hear that the automatic standby works on Linux too, so the battery doesn't go empty constantly.

And you're right, Logitech uses some Blue plugins on Windows. But I discovered that we can get really good audio quality on Linux too by using the EasyEffects app: https://github.com/wwmm/easyeffects

Thanks for everything. You're a fantastic person. :)

PR #209, opened 2022-05-01
feat: add steelseries arctis 7 plus (2021) support

Michsior14:
This adds Steelseries Arctis 7 Plus support for side tone and inactive time (tested). 
~I couldn't find in wireshark any info regarding battery status so this is skipped for now. I would welcome help though in this matter.~

Additionally this model support a few Equalizer presets (also custom settings) that I can successfully change via following commands:

```sh
# flat
./headsetcontrol --dev -- --device 0x1038:0x220e --interface 3 --send "0x0, 0x33, 0x18, 0x18, 0x18, 0x18, 0x18, 0x18, 0x18, 0x18, 0x18, 0x18, 00"

# bass boost
./headsetcontrol --dev -- --device 0x1038:0x220e --interface 3 --send "0x0, 0x33, 0x1f, 0x20, 0x1a, 0x15, 0x15, 0x16, 0x16, 0x16, 0x16, 0x23, 00"

# smily
./headsetcontrol --dev -- --device 0x1038:0x220e --interface 3 --send "0x0, 0x33, 0x1e, 0x1b, 0x15, 0x10, 0x10, 0x13, 0x1b, 0x1e, 0x20, 0x1f, 00"

# focus
./headsetcontrol --dev -- --device 0x1038:0x220e --interface 3 --send "0x0, 0x33, 0x0e, 0x16, 0x11, 0x13, 0x20, 0x24, 0x1f, 0x11, 0x18, 0x11, 00"
```

I can add this to the project if you are interested in supporting EQ.

PR #210, opened 2022-05-02
feat: support equalizer presets for arctis 7+

Michsior14:
As mentioned in PR #209 I've added the support for equalizer presets. 
Not sure if this is the way you wan't to handle this. If something is off please propose the changes.

### DIFF
--- a/src/devices/steelseries_arctis_7_plus.c
+++ b/src/devices/steelseries_arctis_7_plus.c
@@ -1,6 +1,7 @@
 #include "../device.h"
 #include "../utility.h"
 
+#include <bits/stdint-uintn.h>
 #include <hidapi.h>
 #include <stdlib.h>
 #include <string.h>
@@ -20,6 +21,7 @@ static const uint16_t PRODUCT_IDS[] = { ID_ARCTIS_7_PLUS };
 
 static int arctis_7_plus_send_sidetone(hid_device* device_handle, uint8_t num);
 static int arctis_7_plus_send_inactive_time(hid_device* device_handle, uint8_t num);
+static int arctis_7_plus_send_equalizer_preset(hid_device* device_handle, uint8_t num);
 static int arctis_7_plus_request_battery(hid_device* device_handle);
 
 int arctis_7_plus_read_device_status(hid_device* device_handle, unsigned char* data_read);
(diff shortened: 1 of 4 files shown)


Michsior14:
Hey @Sapd, any update on this?

src/devices/steelseries_arctis_7_plus.c
```diff
@@ -91,6 +94,33 @@ static int arctis_7_plus_request_battery(hid_device* device_handle)
...
+        uint8_t smiley[MSG_SIZE] = { 0x0, 0x33, 0x1e, 0x1b, 0x15, 0x10, 0x10, 0x13, 0x1b, 0x1e, 0x20, 0x1f, 0x0 };
+        return hid_write(device_handle, smiley, MSG_SIZE);
+    }
+    case 3: {
+        uint8_t focus[MSG_SIZE] = { 0x0, 0x33, 0x0e, 0x16, 0x11, 0x13, 0x20, 0x24, 0x1f, 0x11, 0x18, 0x11, 0x0 };
+        return hid_write(device_handle, focus, MSG_SIZE);
+    }
+    }
+
+    return 0;
```
Sapd, line 121: instead of the ending `return 0;` we should use a 
`default:` and then inside write a error message that it needs to be between 0 and 4 and return an HSC_ERROR (maybe a new one, like HSC_OUTOFBOUNDS defined in `device.h`).

Reason we need this redundant is, that if other headsets want also define the `-e` setting, we might modify the check in `main.c` to support more possibilities. So the implementation needs to also check for itself if the parameter is defined.

Michsior14: Adjusted according to your recommendation ;)

src/main.c
```diff
@@ -271,6 +285,7 @@ int main(int argc, char* argv[])
         { "chatmix", no_argument, NULL, 'm' },
         { "dev", no_argument, NULL, 0 },
         { "help", no_argument, NULL, 'h' },
+        { "equalizer-preset", required_argument, NULL, 'e' },
```
Michsior14, line 288: @Sapd I am also wondering about the shortcut for this. Maybe it should'be just `p`? I could see `e` be used for custom preset support.

Sapd: Yep good suggestion 

Rest looks good

Michsior14: Updated!

PR #211, opened 2022-05-08
initial support for hyperx cloud flight wireless

hede5562:
initial support for hyperx cloud flight wireless

There are several revisions of this headset but I do only have one. Maybe other revisions need further work but I didn't get any review on my fork.


hede5562:
fixed some problems with the code style... try again...

hede5562:
format-check fails were valid, but I do not see the point for the last whitespace failures as there are no longer any trailing whitespaces?

hede5562:
Thank you for inclusion. Yes, the whitespace check seems to check those also for all intermediate patches. It seems without some kind of a rebase / squashing of those patches these kind of errors won't go away. Well, good to have a human reviewer here which can overrule automatic checks - seems way easier than the rebase/squash thing. ;-)

PR #220, opened 2022-06-01
Add Arctis 7X support

blades:
Added the ID for Arctis 7X to the Arctis 1 class.


blades:
I'll continue to test just in case something crops up.

PR #227, opened 2022-08-15
Moving battery calculation for lipo discharge into utility.c

jaypikay:
I have moved the calculation of the battery level over to `utility.c` and updated the code for the Logitech Pro X to use it. I did not touch the `logitech_g633_g933_935.c` as I cannot verify the correctness, when porting it.

### DIFF
--- a/src/devices/logitech_gpro.c
+++ b/src/devices/logitech_gpro.c
@@ -1,11 +1,13 @@
-#include "../device.h"
-#include "../utility.h"
-#include "logitech.h"
-
 #include <unistd.h>
 
 #include <math.h>
 #include <string.h>
+#include <stdint.h>
+
+#include "../device.h"
+#include "../utility.h"
+#include "logitech.h"
+
 
 static struct device device_gpro;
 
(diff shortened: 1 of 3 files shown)


Sapd:
Thank you, its a good PR for cleaning up.

However we should also provide a function for simple splines. As polynomials are a bit of overkill (and more complicated) for most cases, i.e.:

```
static float estimate_battery_level(uint16_t voltage) {
    int p[] = { 100, 60, 10, 0 };
    int v[] = { 4200, 3870, 3690, 3270 };
    int size = 4;
    
    float percent = 0;
    for(int i = 0; i < size; ++i) {
        // if >= then 100%
        if(voltage >= v[i]) {
            percent = p[i];
            break;
        } 
        
        // if not last
        if(i < size-1 && voltage >= v[i+1]) {
            percent = p[i+1] + (voltage-v[i+1])/(v[i]-v[i+1])*(p[i]-p[i+1]);
            break;
        }
       return percent;
}
```

We maybe should also start writing tests for utility functions.

PR #230, opened 2022-08-30
Add a monitor mode

hadess:
Repeat the commands requested after sleeping 2 seconds. This can be
useful to monitor a battery status.


Sapd:
I would additionally suggest, that the monitor flag has an optional parameter, which defines how many seconds, defaulting to 2.

hadess:
> I would additionally suggest, that the monitor flag has an optional parameter, which defines how many seconds, defaulting to 2.

Got any recommendations for the name of that option?

PR #251, opened 2022-11-12
feat: add arctis 7+ chatmix

Michsior14:
Add chatmix support for 7+ based on arctis nova implementation which turned out to work well.


Azarilh:
I am confused, how is Chatmix supposed to work? I have the Arctis 7+ which are supposedly compatible with Chatmix, and i do get a value that changes depending on what i set it to from my headphones, but i hear no difference. From my understanding it changes the volumes of the gaming and chat output virtual devices, but i only have one.

PR #254, opened 2022-11-13
feat: add custom equalizer to arctis 7+ and nova 7

Michsior14:
Add custom equalizer bands values support and implement it for arctis 7+ and nova 7.

Resolves #253


Sapd:
Good work!

PR #262, opened 2022-12-31
Adds RGB control feature

mateocervilla:
## Description

This pull request adds the arguments required for setting RGB values on some headsets and the implementation for the Logitech G733.
This development was carried out on a Logitech G733 on Linux.

## Configuration options

```
headsetcontrol -R Sets Red channel light color, values must be between 0 and 255

headsetcontrol -G Sets Green channel light color, values must be between 0 and 255

headsetcontrol -B Sets Blue channel light color, values must be between 0 and 255

headsetcontrol -t 0|1 Sets top light color. Use it with -R -G -B. Default: Bottom (0 = bottom, 1 = top)
```

## Examples
- Set bottom rgb light to purple:

`headsetcontrol -R 255 -B 255`

- Set top rgb light to green:

`headsetcontrol -G 255 -t`

## Clarifications
- RGB colors are scaled. I had to get every value from 0 to 255 and fit a polynomial line. So in the code it was necessary to use a polynomial function to set the final values.
- This function was implemented on the file for the Logitech devices {G633, G635, G933, G935, G733} but only tested on a G733.
- It is possible that other devices may not use the -t (top) argument or may set more than two lights, so when the time comes, it must be changed. Perhaps a light index could be instead.


Sapd:
If all color values are required anyway (rgb), it might make sense to put them into one argument, like --colors 255,255,255 (like on the equaliser). But Im open for discussion here.

Im not sure how the polynomial function was created, but you also might consider using splines, which are simpler and more transparent (if applicable in this case): https://github.com/Sapd/HeadsetControl/pull/197#issuecomment-1006922692

mateocervilla:
> it might make sense to put them into one argument, like --colors 255,255,255 (like on the equaliser). But Im open for discussion here.

That's a good idea actually. I had also thought about setting the colors with their hex codes (--colors FF55AA)


> how the polynomial function was created

I send from the Windows Logitech G Hub all the values from 0 to 255 (changing one color channel) and save every hex value from wireshark. Then I created a table with its respective graph and I calculated a curve that fits the points.

I did something like this guy did (which also has an interesting repo): 

- ["Colors are RGB value but these are scaled. I used polynomial trendline feature on excel and ended up with this"](https://github.com/YulCmr/G733_windows_app#some-reverse)

PR #264, opened 2023-01-07
Add Corsair VOID RGB USB 3

rwjack:
Not sure why, but there seems to be another product ID for these headphones. Mine are white so it could be the black/yellow/white combo we have here.


stuarthayhurst:
@rwjack Do the sidetone and alert options work for you? I've got a different headset that's a little quirky, just trying to figure out if it's my model only or all wired headsets.

rwjack:
I don't really remember, sold my pair about a year ago, wasn't using them on Linux anyways.

stuarthayhurst:
No worries, thanks anyway

PR #266, opened 2023-01-17
Improve Logitech G535 support

Moonlight-Angel:
Building upon #234, this will improve the Logitech G535 support by adding battery status and inactive time.

### DIFF
--- a/src/devices/logitech_g535.c
+++ b/src/devices/logitech_g535.c
@@ -4,14 +4,79 @@
 
 #include <math.h>
 #include <string.h>
-
-#define MSG_SIZE 20
+#include <unistd.h>
 
 static struct device device_g535;
 
 static const uint16_t PRODUCT_ID = 0x0ac4;
 
+// Based on manual measurements so the discharge curve used to generate these values aren't always
+// right, but it's good enough.
+// Based on the following measured values on a brand new headset (voltage, percentage) :
+// - 4175, 100
+// - 4135, 98
+// - 4124, 97
+// - 4109, 96
+// - 4106, 95
+// - 4066, 90
+// - 4055, 87
+// - 4047, 86
+// - 4036, 85
+// - 4025, 84
+// - 4000, 83
+// - 3985, 81
+// - 3974, 80
+// - 3971, 79
+// - 3963, 78
+// - 3945, 72
+// - 3934, 71
+// - 3916, 67
+// - 3894, 64
+// - 3887, 63
+// - 3872, 61
+// - 3839, 56
+// - 3817, 50
+// - 3806, 48
+// - 3788, 39
+// - 3774, 34
+// - 3766, 30
+// - 3752, 26
+// - 3741, 22
+// - 3730, 20
+// - 3719, 17
+// - 3701, 13
+// - 3688, 10
+// - 3679, 8
+// - 3675, 6
+// - 3664, 5
+// - 3640, 4
+// - 3600, 3
+// - 3540, 2
+// - 3485, 1
+// - 3445, 1
+// - 3405, 1
+// - 3339, 0
+// - 3325, 0
+// - 3310, 0
+
+// Generated using https://arachnoid.com/polysolve/
+static const double battery_estimate_terms[] = {
+    5.2745819811026892e+006,
+    -7.2266407634766219e+003,
+    3.8767801227503904e+000,
+    -1.0542664248696289e-003,
+    1.8582262783020881e-007,
+    -3.4500263384672060e-011,
+    4.7875908463303429e-015,
+    3.9178868212079191e-019,
+    -2.1037922411043121e-022,
+    1.7519707367951941e-026
+};
+static const size_t num_terms = 10;
+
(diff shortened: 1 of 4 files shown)


src/devices/logitech_g535.c
```diff
@@ -4,14 +4,79 @@
...
+#include <unistd.h>
 
 static struct device device_g535;
 
 static const uint16_t PRODUCT_ID = 0x0ac4;
 
+// Based on manual measurements so the discharge curve used to generate these values aren't always
+// right, but it's good enough.
+// Based on the following measured values on a brand new headset (voltage, percentage) :
+// - 4175, 100
```
Moonlight-Angel, line 16: This takes a lot of space but I think it's useful information if somebody else wants to implement the discharge curve or battery status in another way.
Maybe put it in another file somewhere else, a CSV file for example ? :shrug:

src/devices/logitech_g535.c
```diff
@@ -21,21 +86,121 @@ void g535_init(struct device** device)
...
+
+    // actual voltage is byte 4 and byte 5 combined together
+    const uint16_t voltage = (buf[4] << 8) | buf[5];
+
+    return (int)(roundf(poly_battery_level(battery_estimate_terms, num_terms, voltage)));
+}
+
+static int g535_send_inactive_time(hid_device* device_handle, uint8_t num)
+{
+    // Accepted values are 0 (never), 1, 2, 5, 10, 15, 30
```
Moonlight-Angel, line 177: Don't know how to limit the given value here, if a value out of these bounds is given, the headset apparently sets the inactive time to "never"

src/devices/logitech_g535.c
```diff
@@ -21,21 +86,121 @@ void g535_init(struct device** device)
...
 }
 
 static int g535_send_sidetone(hid_device* device_handle, uint8_t num)
 {
+    int ret = 0;
+
     num = map(num, 0, 128, 0, 100);
 
-    uint8_t set_sidetone_level[MSG_SIZE] = { 0x11, 0xff, 0x04, 0x1e, num };
+    uint8_t buf[HIDPP_LONG_MESSAGE_LENGTH] = { HIDPP_LONG_MESSAGE, HIDPP_DEVICE_RECEIVER, 0x04, 0x1d, num, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 };
```
Moonlight-Angel, line 108: The 4th byte here changed from `0x1e` to `0x1d`. Don't know why but `0x1d` is the value I saw on my side when setting the sidetone level

src/devices/logitech_g535.c
Sapd: I think instead of a polynomial function its better to have a spline. A spline is much simpler and is more transparent in defining it. 

here is sample code for  that: https://github.com/Sapd/HeadsetControl/pull/197#issuecomment-1006922692

it btw. does not need to the that accurate, as every battery will be different anyway, even of the same headset

src/devices/logitech_g535.c
Sapd: You can return return HSC_OUT_OF_BOUNDS; then

src/devices/logitech_g535.c
```diff
@@ -163,6 +163,10 @@ static int g535_request_battery(hid_device* device_handle)
 static int g535_send_inactive_time(hid_device* device_handle, uint8_t num)
 {
     // Accepted values are 0 (never), 1, 2, 5, 10, 15, 30
+    if (num != 0 && num != 1 && num != 2 && num != 5 && num != 10 && num != 15 && num != 30) {
+        return HSC_OUT_OF_BOUNDS;
+    }
```
Moonlight-Angel, line 168: Don't know if there is a cleaner way to check if `num` is one of the accepted values in C, normally I'd check if `num` is in an array of the accepted values but that looks very complicated to do in C :joy:

src/devices/logitech_g535.c
Moonlight-Angel: I've implemented a utility function to use splines to compute the battery level :+1: 
I'd still keep all the measured values just for documentation

19Gerhard85:
When will this be merged?

Moonlight-Angel:
> When will this be merged?

Waiting for the second round of reviews, no ETA

src/devices/logitech_g535.c
Sapd: I will keep that in; we can then see if there are complaints :P

src/devices/logitech_g535.c
Sapd: I Will improve that today. It needs an error message with the correct values and a good way would be to round the user supplied value to multiple of 5 for num > 3 and then to use that. So that the user does not have to look up the correct values before that (except of maximum)

Sapd:
> When will this be merged?

Will merge it today, after I made a small improvement in the out of bounds check

PR #283, opened 2023-04-13
Graphical User Interface

JoePfeiffer:
This is an initial try at a GUI for headsetcontrol, written in python and GTK.

It supports all of the program options provided by headsetcontrol; after querying the headset for capabilities it constructs a window allowing modification and monitoring of those capabilities:

- capabilities that require a string as input are mapped to a Gtk.Entry (equalizer string)
- on/off capabilities are mapped to a Gtk.Switch (light, voice-prompt, rotate-to-mute)
- capabilities allowing a small number of choices are mapped to a Gtk.SpinButton (equalizer-preset)
- capabilities allowing many choices are mapped to a Gtk.Scale. (sidetone, inactive-time)
- capabilities for monitoring are mapped to a Gtk.ProgressBar (battery, chatmix)

I only have a G533, so only the sidetone, inactive-time, and battery capabilities are fully tested

Turning on debug with -d with bring up the GUI window with all capabilities shown and active; attempts to use capabilities that are not supported by the headset will print the command arguments but will not attempt to actually send the command to the headset.

### DIFF
--- a/python/headsetcontrol-gui.py
+++ b/python/headsetcontrol-gui.py
@@ -14,7 +14,7 @@ try :
 except :
     sys.exit("import failed: please ensure that you have the gi (Python API for Introspection) library installed")
 
-# GUI
+# GUI -- mainwindow
 class MainWindow(Gtk.ApplicationWindow):
     def quit(self, button) :
         Gtk.main_quit()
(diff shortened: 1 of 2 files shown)


Sapd:
Thank you for your contribution. I appreciate the effort and time you have put into. However, I believe it would be best to create a new repository for the GUI rather than merging it directly into HeadsetControl.

Thats for multiple reasons:
* Merging in code means, that it must be maintained by the main repository (in this case by me)
* The UNIX philosophy advocates for a modular software approach (where each tool does one thing well).  By seperating software on top into different repositories, this adheres to this principle, as each tool can focus on its specific task without being encumbered by additional code that is not relevant to its purpose
* GUI tools already exist in seperate repositories, this way multiple ones can exist to be chosen from: https://github.com/Sapd/HeadsetControl#third-party


Feel free to put your python code in a separate repository, I would suggest putting it in a completely separate structure without the source code of this repo. You can then create a PR here linking to yours in the Readme (like done with the other software, see link above)

Issue #277, opened 2023-03-19
Steelseries Nova Pro Wireless (New revision)


Eb-:
### Description

Picked up a Nova Pro Wireless, unfortunately it looks like it's a new revision that isn't recognized by `headsetcontrol`. Looks like this one has a ProductID of 0x12e5 instead of the expected 0x12e0.

### Headset Name

Steelseries Nova Pro Wireless

### Device information

<details>
 <summary>Detailed Device Information</summary>

 ```
 $ headsetcontrol --dev -- --list --device 0x1038:0x12e5
Device Found
  VendorID: 0x1038
 ProductID: 0x12e5
  path: /dev/hidraw16
  serial_number: 
  Manufacturer: SteelSeries
  Product:      Arctis Nova Pro Wireless
  Interface:    3
  Usage-Page: 0xc Usageid: 0x1

Device Found
  VendorID: 0x1038
 ProductID: 0x12e5
  path: /dev/hidraw17
  serial_number: 
  Manufacturer: SteelSeries
  Product:      Arctis Nova Pro Wireless
  Interface:    4
  Usage-Page: 0xffc0 Usageid: 0x1

Device Found
  VendorID: 0x1038
 ProductID: 0x12e5
  path: /dev/hidraw17
  serial_number: 
  Manufacturer: SteelSeries
  Product:      Arctis Nova Pro Wireless
  Interface:    4
  Usage-Page: 0xff00 Usageid: 0x1
 ```

</details>

Eb-:
For reference a request for a different revision of the Nova Pro Wireless is found here: https://github.com/Sapd/HeadsetControl/issues/221

Sapd:
Best thing is to try to add the id yourself and then make a PR. As only people with this headset are able to test it.

You can do so according to the Wiki of how to add a Corsair device, but instead you simply use the file which belongs to your headset. Otherwise the steps are similar: https://github.com/Sapd/HeadsetControl/wiki/Adding-a-Corsair-device

Issue #221, opened 2022-07-21
Device Request: Arctis Nova Pro Wireless


Mulugruntz:
### Description

I just received mine. I'll try and see after work if I can follow the steps from your Wiki and find the proper values.

### Headset Name

Arctis Nova Pro Wireless

### Device information

<details>
 <summary>Detailed Device Information</summary>

 ```
$ headsetcontrol --dev -- --list --device 0x1038:0x12e0
Device Found
  VendorID: 0x1038
 ProductID: 0x12e0
  path: /dev/hidraw7
  serial_number: 
  Manufacturer: SteelSeries
  Product:      Arctis Nova Pro Wireless
  Interface:    3
  Usage-Page: 0xc Usageid: 0x1

Device Found
  VendorID: 0x1038
 ProductID: 0x12e0
  path: /dev/hidraw8
  serial_number: 
  Manufacturer: SteelSeries
  Product:      Arctis Nova Pro Wireless
  Interface:    4
  Usage-Page: 0xffc0 Usageid: 0x1

Device Found
  VendorID: 0x1038
 ProductID: 0x12e0
  path: /dev/hidraw8
  serial_number: 
  Manufacturer: SteelSeries
  Product:      Arctis Nova Pro Wireless
  Interface:    4
  Usage-Page: 0xff00 Usageid: 0x1

 ```

</details>

Mulugruntz:
When changing the Sidetone in Steelseries GG -> Engine.
![image](https://user-images.githubusercontent.com/13810199/180641867-7a3b739a-a6c0-4369-b3a4-31f275a917ae.png)

Sidetone `OFF`:
```
0000   1b 00 c0 57 e9 b6 01 e6 ff ff 00 00 00 00 09 00   ...W............
0010   00 02 00 13 00 04 01 40 00 00 00 06 39 *00* 00 00   .......@....9...
0020   00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00   ................
0030   00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00   ................
0040   00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00   ................
0050   00 00 00 00 00 00 00 00 00 00 00                  ...........
```

Sidetone `LOW`:
```
0000   1b 00 c0 b7 6f b4 01 e6 ff ff 00 00 00 00 09 00   ....o...........
[...]

PR #306, opened 2023-09-20
Add Steelseries Arctis Nova Pro Wireless support

Anden:
Added some support for Steelseries Arctis Nova Pro Wireless.

Equalizers and light (mic mute led) should be possible to control as well, will try to add that in the coming days. But if life gets in the way this might be enough for a start.


Anden:
Should be good to merge now.

One thing to note is that this is tested on the PC/PS5 version.
Do you want me to add the 0x12e5 id that seems to be for the PC/XBOX version (reported in #221 and #277)? It will probably work without changes, but I've no way of verifying that.

PR #307, opened 2023-10-07
Support for 0x0a56(Corsair Void Elite USB RGB)

catdotjs:
While there is a entry for Corsair Void Elite USB. The RGB version seems to have a separate id.


LooksMaxing:
Can you provide both ids?

catdotjs:
It's in the commit

LooksMaxing:
I found those internet articles an a bash script that might help you

https://gist.github.com/pubkey/dd5f239737bb7a2c4dc17847a15fc8a1
https://www.reddit.com/r/linux_gaming/comments/eo92hz/anyone_know_how_to_control_rgb_on_wireless/
https://bbs.archlinux.org/viewtopic.php?id=276722 (if you use arch like me)

Sapd:
Thank you! Sorry for the delay

PR #320, opened 2023-12-27
Introduce internal equalizer baseline and band limit

AnErrupTion:
Hello! This PR introduces an internal equalizer baseline and band limit. On the one hand, having the baseline means is that, instead of having to input band values specific to the device for the equalizer, e.g.:

``headsetcontrol -e "0x1c 0x19 0x11 0x14 0x14 0x14"``

you can specify the band value shown in the software, e.g.:

``headsetcontrol -e "4 2.5 -1.5 0 0 0"``

Essentially, the baseline is the band value sent in a USB command when it's flat. The band limit, on the other hand, sets limits for those arbitrary values in the equalizer argument string. For example, on the Arctis Nova 3, the minimum and maximum band values are ``-6`` and ``+6`` respectively.

Moreover, this PR fixes potential bugs. For example, the send and send-feature commands used a **signed** byte buffer instead of an unsigned one, which could cause some issues.

Furthermore, the function ``get_data_from_parameter`` in ``utility.c`` (now named ``get_byte_data_from_parameter``) allocated a buffer which was never used, but worse, sometimes was never freed if a certain condition was met. That buffer has been removed.

Finally, I've mentionned support for the Arctis Nova 3 on the README file, something I forgot to do in my last PR. 😅

~~However, this PR has been created as a draft PR due to one issue. The Arctis Nova Pro Wireless is yet another headset that implements the EQ feature, however, I've been unable to find its baseline value. This is because the Equalizer Preset capability for this headset directly sends a value corresponding to an internal preset, unlike the others implementing it which basically send a specific equalizer string directly to the headset.
(cc @Anden: If you can help me figure out the baseline value (perhaps by setting a custom all-flat EQ preset), it would be awesome! Thanks!)~~
EDIT: This has been solved now!

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,8 @@ talking. This differs from a simple loopback via PulseAudio as you won't have an
   - Sidetone, Battery
 - SteelSeries Arctis 1, Arctis 1 for XBox
   - Sidetone, Battery, Inactive time
+- SteelSeries Arctis Nova 3
+  - Sidetone, Equalizer Presets, Equalizer, Microphone Mute LED Brightness, Microphone Volume
 - SteelSeries Arctis (7 and Pro)
   - Sidetone, Battery, Inactive time, Chat-Mix level, LED on/off (allows to turn off the blinking LED on the base-station)
 - SteelSeries Arctis 7+
--- a/src/dev.c
+++ b/src/dev.c
@@ -173,8 +173,8 @@ int dev_main(int argc, char* argv[])
     int print_deviceinfo = 0;
 
 #define BUFFERLENGTH 1024
-    char* sendbuffer       = calloc(BUFFERLENGTH, sizeof(char));
-    char* sendreportbuffer = calloc(BUFFERLENGTH, sizeof(char));
+    unsigned char* sendbuffer       = calloc(BUFFERLENGTH, sizeof(char));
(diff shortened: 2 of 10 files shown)


Anden:
@AnErrupTion The Arctis Nova Pro Wireless base-station allows to set each band from -10 to +10 in 0.5 increments, which corresponds to baseline 20 (0x14), minimum 0 and max 40 (0x28) in the data.

The preset stuff is just to ensure that the custom equalizer takes effect.
In `set_equalizer` the passed band values are set immediately after enabling custom equalizer:
```c
set_equalizer_preset(device_handle, EQUALIZER_PRESET_CUSTOM);

uint8_t data[MSG_SIZE] = { 0x06, 0x33 };
for (int i = 0; i < settings->size; i++) {
    data[i + 2] = (uint8_t)settings->bands_values[i];
}
```

AnErrupTion:
@Anden Thank you! This is exactly what I needed. :D

Anden:
I tested with my Arctis Nova Pro Wireless, and it works as expected. :+1:

AnErrupTion:
@Sapd Just so you know, this is ready to be reviewed/merged. 😄

Sapd:
Thank you!

Issue #322, opened 2024-01-27
Add option to check if headset is connected


AinoSpring:
### Description

When writing scripts using headsetcontrol, it would be very practical having
a flag which checks if there is a headset connected.

Sapd:
Will be provided indirectly over 
https://github.com/Sapd/HeadsetControl/issues/271
(Feel free to comment there for feedback)

AinoSpring:
Yeah but when writing stuff like bash scripts (e. g. for an i3 status bar module) it
would be really convenient if you could do something like

`if [ headsetcontrol --connected ] ; then`

and it feels like it should exist.

Sapd:
Makes sense, will probably implement that and --count or so to check how many are connected (used for further use cases).

PR #323, opened 2024-01-28
Closes issue #322 | Add option to check if headset is connected

AinoSpring:
This closes issue #322. It adds the command line flag `--connected` alias `-t` (standing for test)
to check if there is a headset connected.

It works by first checking if a headset is detected. 
If it does, there are 2 cases:
- a headset is connected and turned on
- a wireless headset's usb is connected but it is turned off

These 2 cases are differentiated by first checking if the
model has the battery status capability.

If it doesn't, it is wired connected.

If it does, the battery status is checked. If an error is returned, the headset
is turned off.

For example, this option can be integrated in bash scripts using
`if headsetcontrol -t ; then ; <do something> ; fi`

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -295,6 +295,7 @@ int main(int argc, char* argv[])
     int sidetone_loudness                = -1;
     int request_battery                  = 0;
     int request_chatmix                  = 0;
+    int request_connected                = 0;
     int notification_sound               = -1;
     int lights                           = -1;
     int inactive_time                    = -1;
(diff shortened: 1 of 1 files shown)


AinoSpring:
There is also true being returned when a headset
is connected making it compatible with the
`test` command.

Sapd:
Thank you! However I would remove the `-t` short option and keep it for other purposes (given --connected is just used for scripting). To do so you basically set it to NULL and handle it in `case 0:` 

I can also update the code but will not get to do so before next week

AinoSpring:
That makes sense. I just implemented and pushed it.

Issue #271, opened 2023-02-16
Output format in kv format


devnore:
### Description

I would love to have the option to output in key-value format. This would make life easier when using the output for scripting and not making multiple calls to get different information.

Example (yaml):
```yaml
status: "found"
model: "SteelSeries Arctis (7/Pro)"
battery_percent: "43" 
```

Another example could be to output in a format that could be sourced.

```bash
headsetcontrol_status:="found"
headsetcontrol_model="SteelSeries Arctis (7/Pro)"
headsetcontrol_battery_percent="43" 
```

Thanks for an awesome tool!

Sapd:
It's actually a very good idea and probably much better than the short option right now.

However it would need a bit of work. I would prefer a separate file which receives an array, and which can convert this to both  YAML, JSON or ENV variables depending on preference. Similar to the `-o` option of kubectl and most Go programs.

devnore:
> It's actually a very good idea and probably much better than the short option right now.
> 
> However it would need a bit of work. I would prefer a separate file which receives an array, and which can convert this to both YAML, JSON or ENV variables depending on preference. Similar to the `-o` option of kubectl and most Go programs.

Had the same thoughts. I was aiming to keep it as simple as possible and at the same time not introducing new problems. The only reason why I didn't give JSON as an example it that it is a PITA to write compared to YAML :)

The absolute best solution would be to have multiple formatters (JSON/YAML/TAML/INI/SH) but implementing that would require some rewrites.

Sapd:
I took a look again at which responses we have inside the code. And actually we don't have any nested or dynamic responses, also everything is already there in main.c . That means implementing that feature (even with support of multiple languages) could be quite easy.

However, we would need to define how responses should look like. It should be simple and easy to parse.

I think at the beginning I would support **YAML**, **Sourceable/ENV format**, **JSON**.

Thats what I would purpose (example given in JSON):

### Successful response for command

Will be responded when setting Sidetone. But also when nothing is specified, so that the program can retrieve the number of the connected headset. When nothing is specified, we could also additionally include the capabilities (see below)

```json
{
    "status": "success",
    "device": "Corsair Void",
    "idVendor": "0x1b1c",
    "idProduct": "0x1b27",
}
```

### Successful request

(device etc. will be also returned, just left out here for the discussion)

```json
{
    "status": "success",
    "battery": 42
}
```

Also instead of the number BATTERY_CHARGING could be the status. Should we simply make it like this:

```json
{
[...]

PR #332, opened 2024-03-16
JSON-output

AinoSpring:
So apparently this issue for key-value-outputting (#271) exists for some time so I just went and implemented it.
It uses json because it is mostly used in such applications (rather than yaml or similar mentioned in the issue).

The output may look something like this:

```
$ ./headsetcontrol -jb?
{"device":"Logitech G533","battery":74,"capabilities":["sidetone","battery status","inactive time"]}
```

The system is made for allowing many changes in the future (other output values can be added in as little as one line).
For now, a success/error message directly in json is probably redundant because that is mostly managed by the return codes.


Sapd:
Hey,
there is already a full implementation in the feature-output branch - including a full revamp (as discussed in the issue you linked): https://github.com/Sapd/HeadsetControl/tree/feature-output

PR #380, opened 2024-11-25
Add CI workflow to publish a continuous release

nicola02nb:
### Changes made

Created a CI workflow files that compiles and publish a "Continuos build" on every push to `master` branch.

This gives users a way to get the latest features without building the project on their own

You can check the result on [my fork](https://github.com/nicola02nb/HeadsetControl)

### Checklist

### DIFF
--- a/.github/workflows/continuos-release.yml
+++ b/.github/workflows/continuos-release.yml
@@ -1,0 +1,99 @@
+name: Continuos build workflow
+
+on:
+  workflow_dispatch:
+    inputs: {}
+  push:
+    branches: [master]
+    paths:
+      - "cmake_modules/**"
+      - "src/**"
+      - "CMakeLists.txt"
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    defaults:
+      run:
+        shell: ${{ matrix.shell }}
+    strategy:
+      matrix:
+        os: [windows-latest, ubuntu-latest, macos-latest]
+        include:
+          - os: windows-latest
+            os_name: windows
+            architecture: x86_64
+            shell: msys2 {0}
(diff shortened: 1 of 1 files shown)


Sapd:
Thank you! Great contribution.

I thought also of doing something like this. Currently Windows Builds are in fact pushed but hidden inside the artifcats window.

Just two points:
- Would it make sense to include it already in the existing flows? I myself guess not bc. this one is just excluded on push. Just something to think about
- Does the Release automatically replace other releases named "continous"? 

Also there could be a question about signing releases (as there was before from AUR contributors), but I am not sure how it would work in a flow as that. Currently the normal releases are manually signed by myself.

nicola02nb:
It shouldn't replace latest release, it just creates a new `pre-release`, tagged as continuous

nicola02nb:
> * Would it make sense to include it already in the existing flows? I myself guess not bc. this one is just excluded on push. Just something to think about

In my opinion makes sense so users have an easier way to access to latest features before release wihout needing to search through CI arctifacts...

nicola02nb:
> Also there could be a question about signing releases (as there was before from AUR contributors), but I am not sure how it would work in a flow as that. Currently the normal releases are manually signed by myself.

I dont know much about signing releases... I can only say that my idea was ment only for `pre-release` which should not need signing? I guess you can do it when you create the official release?

Sapd:
> In my opinion makes sense so users have an easier way to access to latest features before release wihout needing to search through CI arctifacts...

I agree

> I dont know much about signing releases... I can only say that my idea was ment only for pre-release which should not need signing? I guess you can do it when you create the official release?

I think that is fine, it could be added afterwards when its requested and really needed

PR #417, opened 2025-08-05
Feat: Add support for SteelSeries Arctis Nova 3P Wireless

sus-domesticus:
I'll be adding support for SteelSeries Arctis Nova 3P Wireless. Right now I managed to get the sidetone working.

I first tried seeing if the implementation of SteelSeries Arctis Nova 3 works for this model as well, but it doesn't.

### DIFF
--- a/src/device_registry.c
+++ b/src/device_registry.c
@@ -22,6 +22,7 @@
 #include "devices/steelseries_arctis_7_plus.h"
 #include "devices/steelseries_arctis_9.h"
 #include "devices/steelseries_arctis_nova_3.h"
+#include "devices/steelseries_arctis_nova_3p_wireless.h"
 #include "devices/steelseries_arctis_nova_5.h"
 #include "devices/steelseries_arctis_nova_7.h"
 #include "devices/steelseries_arctis_nova_pro_wireless.h"
@@ -68,6 +69,7 @@ void init_devices()
     add_device(elo71USB_init);
     // SteelSeries
     add_device(arctis_nova_3_init);
+    add_device(arctis_nova_3p_wireless_init);
     add_device(arctis_nova_5_init);
     add_device(arctis_nova_7_init);
     add_device(arctis_7_plus_init);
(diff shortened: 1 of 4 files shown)


AnErrupTion:
> I first tried seeing if the implementation of SteelSeries Arctis Nova 3 works for this model as well, but it doesn't.

It seems like most (if not all) of the commands for each capability are the same, except without the `0x06` byte at the beginning. So you might be able to copy most of the code from the Arctis Nova 3 implementation for the Nova 3P (though I already saw the latter was a bit more capable, such as supporting more sidetone levels).

marcg03:
Yeah, I copied most of the code

marcg03:
Support for Linux works fine on my end.

Unfortunately inside my Windows VM things don't work (I tried all `Usage-Page` + `Interface` combos as reported by `headsetcontrol --dev -- --list`). There seems to be an issue with how I use `hid_send_feature_report` because inside WireShark I see the packets have 1060 length.

Sapd:
> Support for Linux works fine on my end.
> 
> Unfortunately inside my Windows VM things don't work (I tried all `Usage-Page` + `Interface` combos as reported by `headsetcontrol --dev -- --list`). There seems to be an issue with how I use `hid_send_feature_report` because inside WireShark I see the packets have 1060 length.

Windows support is a bit tricky, it expects the exact same length.
Here it seems to be overwritten:
The FeatureReport seems to be supposed to be longer

https://github.com/libusb/hidapi/blob/f18d2c0768bed300d13758dd9f52b57163c0498f/windows/hid.c#L1256

I think we can merge anyways, we just need to maybe include some message or at least a comment that windows seems to not quite work yet with that implementation

marcg03:
Alright, where should I write the message/comment?

Sapd:
Are all functions erroring out or only specific?

marcg03:
From what I recall, nothing works on windows and they don't error out but instead just don't have any effects.

Issue #415, opened 2025-07-17
Increase Equalizer Preset Number


ModoloDev:
### Description

I’d like to propose increasing the number of available equalizer presets.
Currently, presets are limited between 0 and 3, so headsets that support more than 4 presets are capped.

ModoloDev:
Thanks, @bwachtendorf! Nice work!

PR #423, opened 2025-10-08
feat: Decouple equalizer_presets_count from equalizer_presets

bwachtendorf:
### Changes made

I decoupled equalizer_presets_count from equalizer_presets, so I could select all profiles on my Audeze Maxwell headset.
The only downsides of this implementation (as I can see right now) are:

- input validation of `--equalizer-preset` is open-ended now
- requires "hardcoding"  equalizer_presets_count for each device with preset support

This would also solve #415.

An alternative to "hardcoding" equalizer_presets_count would be to implement dummy presets for Audeze Maxwell and count those like it was done until now.

### Checklist

### DIFF
--- a/src/devices/headsetcontrol_test.c
+++ b/src/devices/headsetcontrol_test.c
@@ -158,7 +158,7 @@ static int headsetcontrol_test_lights(hid_device* device_handle, uint8_t on)
 
 static int headsetcontrol_test_send_equalizer_preset(hid_device* device_handle, uint8_t num)
 {
-    if (num < 0 || num > EQUALIZER_PRESETS_COUNT-1) {
+    if (num < 0 || num > EQUALIZER_PRESETS_COUNT - 1) {
         return -1;
     }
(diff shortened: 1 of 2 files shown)


Sapd:
Ah I think I rebased and you pushed something at the same time.

Note that your profile is not linked in the commits because you probably dont have that email address in the commit in the GitHub profile

Sapd:
From my side it's ready. Tell me if its fine from your side to also merge.

bwachtendorf:
I did not push anything. I was about to push the linting fixes but you were faster than me.

I understand now how I could have implemented the input validation, but at the time I could not find a suitable place to put it. As for the warning, I wanted to fix it before making the PR but forgot.

Otherwise it looks good to me and still works with my Audeze Maxwell headset.

Sapd:
> I did not push anything. I was about to push the linting fixes but you were faster than me.

Thank you! Was probably misreading something during push.

PR #426, opened 2025-10-25
Add support for arctics nova 7 wireless gen 2

sarpilgaz:
### Changes made

Added (partial?) support for Steelseries Arctics nova 7 wireless gen 2. This was done by following the guide on "Adding the deviceid to the implementation" under "Development 1 ‐ Adding a device". Simply this change added support for, as far as I have tested, sidetone and battery.

Only tested the battery and sidetone commands, therefore I have only marked those two in the README file.


Sapd:
Manually merged in 311bd7e498a7d32d89ef844b031956edb9f8322e

PR #429, opened 2025-10-29
Modernize CI: Upgrade to lukka/run-cmake@v10 with CMakePresets.json

Sapd:
Replaces the outdated lukka/run-cmake@v3 with v10


Sapd:
superseded

PR #456, opened 2026-01-10
Add top level project guard

billtran1632001:
Since the recommended way to use the headsetcontrol library is as a subdirectory. Some configuring steps should be disabled. This PR adds a top level project guard to disable some configuring steps that are unnecessary when using as a subdirectory or with [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html).

This leverages [`PROJECT_IS_TOP_LEVEL`](https://cmake.org/cmake/help/latest/variable/PROJECT_IS_TOP_LEVEL.html) variable, which is available since CMake's 3.21. With older version, we fall back to check whether `CMAKE_SOURCE_DIR` is equal to `PROJECT_SOURCE_DIR`.

Potential problem: In this patch I also put the build CLI executable step and generating udev rules step behind the guard. Should generating udev contents also be available through API and let the library consumers handle it themselves? Or the generating udev rules step should not be put behind the guard?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -266,29 +266,29 @@ endif()
 # CLI Executable
 # ------------------------------------------------------------------------------
 
-# Create the CLI executable that links against the library
-# On Windows, include the resource file for the application icon
-if(WIN32)
-    set(WIN_RESOURCES ${CMAKE_CURRENT_SOURCE_DIR}/assets/headsetcontrol.rc)
-    # Ensure resource compiler can find the icon file
-    set(CMAKE_RC_FLAGS "${CMAKE_RC_FLAGS} -I${CMAKE_CURRENT_SOURCE_DIR}/assets")
-endif()
-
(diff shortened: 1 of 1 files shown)


Sapd:
I think the logic is inverted if I see it right.

Currently with this changes:  
- Building HeadsetControl standalone -> CLI, tests, install, CPack are disabled
- Using as subdirectory → CLI, tests, install, CPack are enabled

So the NOTs have to be removed. Or does it have a specific reason it's this way?

> In this patch I also put the build CLI executable step and generating udev rules step behind the guard. Should generating udev contents also be available through API and let the library consumers handle it themselves? Or the generating udev rules step should not be put behind the guard?

I think it can be as-is for now. We can also later in theory add it into the library

billtran1632001:
> I think the logic is inverted if I see it right.

You're right! Should have not written anything while lacking sleep. 😅

PR #457, opened 2026-01-13
Steelseries Gen2 uses absolute battery

Sapd:


alita8606:
Arctis Nova 7X (old one (gen1)) also got the new battery percentage update

alita8606:
The new PID is 22A5

Issue #475, opened 2026-01-27
SteelSeries Arctis 7's ChatMix is wrong


billtran1632001:
### Description

Correct me if I'm wrong, from the wording of the comment within `ChatmixResult`, a ChatMix value lesser than 64 means that chat volume is decrease, and a value greater than 64 means game volume is decrease instead.

However, on my SteelSeries Arctis 7, it is backward. Turning max game volume reports a value of 127, and turning max chat volume reports 1. (Never reached 128 or 0 on my headset, could be a separate, unrelated physical problem rather than software.)

The calculated percents also backward, too. When sliding toward game volume from neutral, the chat percentage jump from 100% to increasing from 0%, i.e. a ChatMix value of 65 reports game volume percentage as 100% and chat volume percentage as 2%. So for max game volume, the percentages are 100% and 99% for game and chat, respectively. Same behavior when sliding toward the other side.

### On which OS does the problem happen?

Linux

Sapd:
Can you try this https://github.com/Sapd/HeadsetControl/pull/480 ?

PR #443, opened 2025-12-14
C++ Rewrite

Sapd:
Rewrite codebase from C to modern C++20.

A part of it was obviously done using AI Agents, it would have been impossible otherwise.
See docs/ for examples how now the modern library works.

Complete modernization introducing type safety, better error handling, and a cleaner architecture.

- **Language**: C → C++20 (requires GCC 10+, Clang 10+, MSVC 2019+)
- **Structure**: Reorganized into lib/, cli/, tests/
- **Error handling**: Result<T> type with rich error information
- **Device code**: 50-70% reduction via protocol templates

---

- High-level C++ API (headsetcontrol.hpp)
- C API for FFI bindings (headsetcontrol_c.h)
- Shared library support (-DBUILD_SHARED_LIBRARY=ON)
- Protocol templates: HIDPPDevice, SteelSeriesNovaDevice
- Data-driven capability system
- Test suite

---

- HIDDevice base class with virtual methods per capability
- Device registry singleton for device lookup
- Capability descriptors as single source of truth
- Feature handler registry (replaces switch statements)

---

CLI interface unchanged - fully backwards compatible.

- See docs/ADDING_A_DEVICE.md for adding devices
- See docs/ADDING_A_CAPABILITY.md for adding features
- See docs/LIBRARY_USAGE.md for library integration

Fixes #431

PR #480, opened 2026-02-05
Fix inverted ChatMix level for SteelSeries Arctis 7/Pro (#475)

Sapd:
The chatmix level calculation had game and chat directions swapped: max game volume produced level > 64 (indicating chat), and vice versa.

This bug originated in the old C implementation (steelseries_arctis_7.c) and was carried over verbatim during the C++ rewrite (#443). It went unnoticed because the old code returned a plain int with no defined semantics for direction, users just saw a changing number. The bug became observable once ChatmixResult defined < 64 as game and > 64 as chat.

### DIFF
--- a/lib/devices/steelseries_arctis_7.hpp
+++ b/lib/devices/steelseries_arctis_7.hpp
@@ -167,8 +167,8 @@ public:
         }
 
         // Calculate percentages (game/chat are 191-255, neutral at 255)
-        int game_pct = (game == 0) ? 100 : map(game, 191, 255, 100, 0);
-        int chat_pct = (chat == 0) ? 100 : map(chat, 191, 255, 100, 0);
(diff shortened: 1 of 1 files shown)


billtran1632001:
That corrected the level value. But the percentages are still wrong. These lines:

https://github.com/Sapd/HeadsetControl/blob/3785e2872895ce43d20f68561d8856b6e605ef52/lib/devices/steelseries_arctis_7.hpp#L170-L171

Need to be changed to:
```cpp
int game_pct = (game == 0) ? 100 : map(game, 191, 255, 0, 100);
int chat_pct = (chat == 0) ? 100 : map(chat, 191, 255, 0, 100);
```

Sapd:
@TheBill2001  Can you please try again

billtran1632001:
Yep, everything is correct now.

Issue #488, opened 2026-02-17
Logitech G522 LIGHTSPEED support


sfu69:
### Description

I will try gather as much information as possible via the wiki on the weekend.

I've already tried adding the deviceId towards the cpp code in the logitech G535 file with no effect, will try out some others on the weekend first before going the wireshark route.

### Headset Name

Logitech G522 LIGHTSPEED 

### Device information

<details>
 <summary>Detailed Device Information</summary>

 ```
headsetcontrol --dev -- --list

Device Found
  VendorID:  0x046d
  ProductID: 0x0b18
  Path: /dev/hidraw5
  Serial: 0000000000000000
  Manufacturer: Logitech
  Product: G522 LIGHTSPEED - Wireless Mode
  Interface: 3
  Usage-Page: 0x000c  Usage-ID: 0x0001

Device Found
  VendorID:  0x046d
  ProductID: 0x0b18
  Path: /dev/hidraw5
  Serial: 0000000000000000
  Manufacturer: Logitech
  Product: G522 LIGHTSPEED - Wireless Mode
  Interface: 3
  Usage-Page: 0xff13  Usage-ID: 0x0001

Device Found
  VendorID:  0x046d
  ProductID: 0x0b18
  Path: /dev/hidraw5
  Serial: 0000000000000000
  Manufacturer: Logitech
  Product: G522 LIGHTSPEED - Wireless Mode
  Interface: 3
  Usage-Page: 0xffa0  Usage-ID: 0x0001
 ```

</details>

### New Device Implementation Checklist

- [x] I acknowledge that implementing a new device requires direct access to the device and familiarity with capturing and analyzing its communications. I have access to the device and am willing to contribute by capturing necessary data and/or attempting implementation.
- [ ] I have already attempted to capture device communications using WireShark or similar tools and am including my findings (successful or not) with this issue to assist in implementation efforts.
- [x] I am seeking guidance on how to begin this process and am willing to collaborate with the community to add support for this device. I understand where to find resources (refer to the WIKI).

ChrisLauinger77:
#477 seems like the lightspeed series made significant changes ...
Did you try "headsetcontrol -o json" after you updated the ID ?

sfu69:
Same issues likely as in #477 
I did update the firmware this week via the g hub on my macbook.

Tested with LogitechG533 this time. The test with G535 is similar.

```JSON
{
  "name": "HeadsetControl",
  "version": "continuous-23-g8d16611-modified",
[...]

PR #478, opened 2026-01-30
Adds support for a new G PRO X2 variant

ChrisLauinger77:
Implements a dedicated device class for the Logitech G PRO X 2 LIGHTSPEED headset (PID 0x0af7).

This device utilizes a unique vendor-specific 64-byte HID protocol for battery status, sidetone, and inactive time management, distinct from the HID++ protocol used by other Logitech G PRO models. The existing G PRO class description is clarified to reflect its HID++ specific implementation.

Includes comprehensive registry lookup and protocol parsing tests to validate the new device's functionality.

PR #501, opened 2026-03-17
Add support for Logitech G522 Lightspeed

iowi479:
### Changes made

An implementation of the basic features of the Logitech G522 Lightspeed.

As stated in #488, the Lightspeed series made significant changes. 
This is directly based on #478 and the findings about the recent changes with the Lightspeed series.

The protocol used with the G522 seems to be mostly the same as for the G PRO X 2 LIGHTSPEED, as the findings are applicable here.
Just the identifiers used in the protocol are different from those in the G PRO X 2 LIGHTSPEED implementation of #478.

#### Side effects with InactiveTime
The command to set the inactive timer also sets other timers (light changes after inactivity and lights off after inactivity).
These values get overwritten each time the command gets sent.
With the filled zero-bytes, these timers get defaulted to `never`.

### Checklist


ChrisLauinger77:
@iowi479 does sidetone work for you in the G522 ?

PR #506, opened 2026-04-06
Added full support for Audeze Maxwell 2 PS/PC version

MonsterBagels:
### Changes made

I created a new device profile for the Audeze Maxwell 2 headset (specifically the PlayStation/PC version). It is based on the profile for the first Maxwell but required enough changes that it warranted a new profile rather than adding device ID to the existing profile. I also added a new function for the "AI microphone noise filter" that comes with the Maxwell 2 using the wiki as a guide.

Everything functions so far on my Linux install without issue. 

### Checklist

- [X] I adjusted the README (if needed)

### DIFF
--- a/lib/devices/audeze_maxwell2.hpp
+++ b/lib/devices/audeze_maxwell2.hpp
@@ -12,35 +12,34 @@ using namespace std::string_view_literals;
 namespace headsetcontrol {
 
 /**
- * @brief Audeze Maxwell Gaming Headset
+ * @brief Audeze Maxwell 2 Gaming Headset
  *
  * Features:
  * - Sidetone (32 levels)
  * - Inactive time with discrete levels
- * - Volume limiter
  * - Equalizer presets (10 total: 6 default + 4 custom)
  * - Battery status
  * - Chatmix
  * - Voice prompts
+ * - Mic noise filter
  *
  * Special Requirements:
- * - Complex 18-packet initialization sequence
- * - 60ms delay between packets (mimics official software)
+ * - Complex 21-packet initialization sequence
+ * - 52ms delay between packets (mimics official software)
  * - Uses hid_get_input_report for responses
  */
-class AudezeMaxwell : public HIDDevice {
+class AudezeMaxwell2 : public HIDDevice {
 public:
-    static constexpr std::array<uint16_t, 3> SUPPORTED_PRODUCT_IDS {
-        0x4b19, // Maxwell
-        0x4b18, // Maxwell Xbox Dongle
+    static constexpr std::array<uint16_t, 1> SUPPORTED_PRODUCT_IDS {
+        0x4b29 // Maxwell 2 (PlayStation/PC version)
     };
 
     static constexpr int MSG_SIZE  = 62;
     static constexpr int REPORT_ID = 0x06;
-    static constexpr int DELAY_US  = 60000; // 60ms
+    static constexpr int DELAY_US  = 52000; // 52ms
 
-    // 18-packet initialization sequence (12 unique + 5 status + 1 final)
-    static constexpr std::array<std::array<uint8_t, MSG_SIZE>, 13> UNIQUE_REQUESTS { { { 0x6, 0x8, 0x80, 0x5, 0x5a, 0x4, 0x0, 0x1, 0x9, 0x20 },
+    // 21-packet initialization sequence (14 unique + 6 status + 1 final)
+    static constexpr std::array<std::array<uint8_t, MSG_SIZE>, 15> UNIQUE_REQUESTS { { { 0x06, 0x08, 0x80, 0x05, 0x5A, 0x04, 0x00, 0x01, 0x09, 0x20 },
         { 0x06, 0x08, 0x80, 0x05, 0x5A, 0x04, 0x00, 0x01, 0x09, 0x25 },
         { 0x06, 0x07, 0x80, 0x05, 0x5A, 0x03, 0x00, 0x07, 0x1C },
         { 0x06, 0x08, 0x80, 0x05, 0x5A, 0x04, 0x00, 0x01, 0x09, 0x28 },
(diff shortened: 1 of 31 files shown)


lib/devices/audeze_maxwell2.hpp
```diff
@@ -0,0 +1,385 @@
...
+            auto result = sendGetInputReport(device_handle, STATUS_REQUESTS[i], status_buffs[i]);
+            if (!result) {
+                return result.error();
+            }
+        }
+
+        MaxwellStatus status {};
+
+        // Parse battery (from last initialization message)
+        const auto& buff = status_buffs[0];
```
Sapd, line 179: Bug: The comment says "Parse battery (from last initialization message)" but `buff` is set to `status_buffs[0]` (the mic status response), not `init_response`.

`init_response` is populated on line 162 but never actually read. This should probably be:
```cpp
const auto& buff = init_response;
```
Otherwise battery will likely return incorrect data or BATTERY_UNAVAILABLE.

lib/devices/audeze_maxwell2.hpp
```diff
@@ -0,0 +1,385 @@
...
+            }
+
+            // Check microphone status (note: BatteryResult has mic_status field)
+            if (status_buffs[1][12] == 0xFF) {
+                status.battery.mic_status = MICROPHONE_UP;
+            }
+        }
+
+        // Parse chatmix (0-20 range, center at 10)
+        status.chatmix_level = map(status_buffs[3][12], 0, 20, 0, 128);
```
Sapd, line 209: Chatmix is read from `status_buffs[3]` which corresponds to `STATUS_REQUESTS[3]` (the "AINF" request), but `STATUS_REQUESTS[2]` is labeled "GameCat mix". Should this be `status_buffs[2]` instead? Please double-check against actual device responses.

Sapd (commented):
Thanks for adding Maxwell 2 support! The architecture is clean and follows existing patterns well. Some comments: Also btw there is a lot of formatting diff.

lib/devices/audeze_maxwell2.hpp
```diff
@@ -0,0 +1,385 @@
...
+    /**
+     * @brief Get comprehensive device status (battery, chatmix, sidetone, filter)
+     *
+     * Must send all 6 status requests to get consistent information
+     */
+    struct MaxwellStatus {
+        BatteryResult battery;
+        int chatmix_level;
+        int sidetone_level;
+        int filter_level;
```
Sapd: `filter_level` is declared here but never populated in `getDeviceStatus()` and never read anywhere. Dead field should be removed or wired up.

lib/devices/audeze_maxwell2.hpp
```diff
@@ -0,0 +1,385 @@
...
+        }
+
+        return InactiveTimeResult {
+            .minutes     = minutes,
+            .min_minutes = 0,
+            .max_minutes = 255
+        };
+    }
+
+    // It doesn't appear that the Maxwell 2 has a volume limiter. It doesn't appear
```
Sapd: `setVolumeLimiter` is implemented here but `CAP_VOLUME_LIMITER` is not in `getCapabilities()`, so this method is unreachable. The comment even says the Maxwell 2 probably doesn't have it. Suggest removing this dead code the base class already returns `notSupported`.

lib/devices/audeze_maxwell2.hpp
```diff
@@ -0,0 +1,385 @@
...
+
+        // Sidetone is enabled whenever level changes
+        auto result = sendGetInputReport(device_handle, cmd);
+        if (!result) {
+            return result.error();
+        }
+
+        if (mapped == 0) {
+            // Toggle sidetone off
+            std::array<uint8_t, MSG_SIZE> toggle_cmd { 0x06, 0x09, 0x80, 0x05, 0x5A, 0x05, 0x00, 0x82, 0x2C, 0x07, 0x00, 0x00 };
```
Sapd, line 249: Minor: the sidetone toggle-off command here sends `0x00` as byte 11, while Maxwell 1 sends `0x01` in the same position. Is this an intentional protocol difference? Worth a brief comment if so.

lib/devices/audeze_maxwell2.hpp
MonsterBagels: Strange it was working yesterday but now it isn't, I'll dig deeper.

lib/devices/audeze_maxwell2.hpp
MonsterBagels: That was a typo on my part, status_buffs[3] is the chatmix status and responds to inputs from the headset. I included a comment with the status_buffs structure to better define it.

lib/devices/audeze_maxwell2.hpp
MonsterBagels: removed.

lib/devices/audeze_maxwell2.hpp
MonsterBagels: Removed.

lib/devices/audeze_maxwell2.hpp
MonsterBagels: I found that with the Maxwell 2 at least, when 0x01 is in byte 11 the headset says "sidetone on" when you command it to turn sidetone off. Changing it to 0x00 fixes it to say "sidetone off" when you turn it off.

lib/devices/audeze_maxwell2.hpp
MonsterBagels: Ok I found my mistake. The battery status is showing up in status_buffs[0] so I removed init_response and fixed the getDeviceStatus function. Battery status is being reported consistently now and agrees with what I'm seeing in the Audeze app in Windows.

MonsterBagels:
Thanks for the review! I pushed the changes in response to your comments. I also ran clang so hopefully the formatting issues have been resolved.

lib/devices/audeze_maxwell2.hpp
```diff
@@ -0,0 +1,374 @@
+#pragma once
+
+#include "../result_types.hpp"
+#include "hid_device.hpp"
+#include <array>
+#include <chrono>
+#include <iomanip> // DEBUG
+#include <iostream> // DEBUG
```
Sapd: These debug includes are still here, should be dropped before merge.

lib/devices/audeze_maxwell2.hpp
```diff
@@ -0,0 +1,374 @@
...
+    }
+
+private:
+    /**
+     * @brief Get comprehensive device status (battery, chatmix, sidetone, AINF, filter)
+     *
+     * Must send all 6 status requests to get consistent information
+     */
+    struct MaxwellStatus {
+        BatteryResult battery;
```
Sapd, line 107: `ainf_level` is declared but never written to anywhere in `getDeviceStatus()`. Either wire it up or just remove it.

lib/devices/audeze_maxwell2.hpp
```diff
@@ -0,0 +1,374 @@
...
+        // status_buffs[0] = battery_status
+        // status_buffs[1] = mic_status
+        // status_buffs[2] = eq
+        // status_buffs[3] = chatmix
+        // status_buffs[4] = ainf_status
+        // status_buffs[5] = sidetone_level
+
+        MaxwellStatus status {};
+
+        const auto& buff = status_buffs[0];
```
Sapd, line 179: If `buff[0] != 0x00` but the `D6 0C 00 00` pattern is never found in the loop, `status.battery` stays default-initialized, which I think means it'll report 0% as BATTERY_AVAILABLE (or whatever the default enum value ends up being). Probably want to init `status.battery.status = BATTERY_UNAVAILABLE` and `level_percent = -1` before the loop so the fallthrough is safe.

lib/devices/audeze_maxwell2.hpp
```diff
@@ -0,0 +1,374 @@
...
+        return SidetoneResult {
+            .current_level = level,
+            .min_level     = 0,
+            .max_level     = 128,
+            .device_min    = 0,
+            .device_max    = 31
+        };
+    }
+
+    Result<InactiveTimeResult> setInactiveTime(hid_device* device_handle, uint8_t minutes) override
```
Sapd, line 265: Small thing: `setInactiveTime` reports `max_minutes = 255` in the result but the rounding logic defaults to 360 minutes. Since the parameter is `uint8_t` (max 255), you can never actually hit the 360-minute tier. Might want a comment noting that, or just cap `final_minutes` at 240 and drop the 360 default.

Sapd (commented):
Nice work overall, left a few things I noticed.

lib/devices/audeze_maxwell2.hpp
MonsterBagels: Whoops, removed.

lib/devices/audeze_maxwell2.hpp
MonsterBagels: I made a comment and limited the final_minutes to 240. I did find that the CLI currently limits inactive time inputs to 90 minutes however.

Sapd (commented):
`status_buffs[4]` is still labeled as `ainf_level` in the comment but nothing actually reads it. The struct doesn't have that field either, so it's just fetched and thrown away. Either drop the comment or wire it up if it's supposed to be used for something.

Issue #477, opened 2026-01-29
Logitech G Pro X 2 LIGHTSPEED cannot be detected


funforums:
### Description

Hi, I would like to report that my Logitech G Pro X 2 LIGHTSPEED is not detected:
`headsetcontrol --capabilities` 
`No supported device found`

I installed it through AUR: https://aur.archlinux.org/packages/headsetcontrol-git
headsetcontrol used to work with this headset, but over time i installed a firmware update and headsetcontrol itself had new versions. I am not entirely sure at which point it stopped detecting it.

Might be related to #471 .

Cheers

### Headset Name

Logitech G PRO X 2 LIGHTSPEED

### On which OS does the problem happen?

Linux

### Device information

<details>
 <summary>Detailed Device Information</summary>

 ```
Device Found
  VendorID:  0x046d
  ProductID: 0x0af7
  Path: /dev/hidraw12
  Serial: 0000000000000000
  Manufacturer: Logitech
  Product: PRO X 2 LIGHTSPEED
  Interface: 3
  Usage-Page: 0x000c  Usage-ID: 0x0001

Device Found
  VendorID:  0x046d
  ProductID: 0x0af7
  Path: /dev/hidraw12
  Serial: 0000000000000000
  Manufacturer: Logitech
  Product: PRO X 2 LIGHTSPEED
  Interface: 3
  Usage-Page: 0xff13  Usage-ID: 0x0001

Device Found
  VendorID:  0x046d
  ProductID: 0x0af7
  Path: /dev/hidraw12
  Serial: 0000000000000000
  Manufacturer: Logitech
  Product: PRO X 2 LIGHTSPEED
  Interface: 3
  Usage-Page: 0xffa0  Usage-ID: 0x0001
 ```

</details>

ChrisLauinger77:
#478 
I added the new deviceID - I get some output - the battery is not working though.
And I think the headset suppots equalizer as well - it does not report the capability.
Can U help capturing data ? I got mine last week it had already the updated firmware.
Never had the "firmware update breaks it" issue in the past ...

ChrisLauinger77:
So I also updated the deviceID into the OLD implementation to make sure the firmware update broke it.
And I confirm the firmware is responsible not the cpp-rewrite.

```
{
  "name": "HeadsetControl",
  "version": "3.1.0-36-gabe3ac8",
  "api_version": "1.3",
  "hidapi_version": "0.14.0",
  "device_count": 1,
  "devices": [
[...]

PR #512, opened 2026-04-18
Logitech G PRO X 2 LIGHTSPEED Centurion protocol

ChrisLauinger77:
### Changes made

Changed from raw package to centurion protocol (for existing battery inactive time and sidetone)
Equalizer with presets is implemented / added

### Checklist

- [x ] I adjusted the README (if needed)

### DIFF
--- a/lib/devices/logitech_gpro_x2_lightspeed.hpp
+++ b/lib/devices/logitech_gpro_x2_lightspeed.hpp
@@ -2,6 +2,7 @@
 
 #include "../utility.hpp"
 #include "protocols/logitech_centurion_protocol.hpp"
+#include <algorithm>
 #include <array>
 #include <chrono>
 #include <cmath>
@@ -26,6 +27,7 @@ public:
     static constexpr uint8_t SIDETONE_DEVICE_MAX                               = 100;
     static constexpr uint8_t SIDETONE_MIC_ID                                   = 0x01;
     static constexpr uint8_t PLAYBACK_DIRECTION                                = 0x00;
+    static constexpr uint8_t EQUALIZER_PRESETS_COUNT                           = 5;
 
     constexpr uint16_t getVendorId() const override
     {
@@ -45,7 +47,26 @@ public:
     constexpr int getCapabilities() const override
     {
         return B(CAP_SIDETONE) | B(CAP_BATTERY_STATUS) | B(CAP_INACTIVE_TIME)
+            | B(CAP_EQUALIZER_PRESET)
             | B(CAP_EQUALIZER) | B(CAP_PARAMETRIC_EQUALIZER);
+    }
+
+    uint8_t getEqualizerPresetsCount() const override
+    {
+        return EQUALIZER_PRESETS_COUNT;
+    }
+
+    std::optional<EqualizerPresets> getEqualizerPresets() const override
+    {
+        EqualizerPresets presets;
+        presets.presets = {
+            { "Flat", std::vector<float>(PRESET_FLAT.begin(), PRESET_FLAT.end()) },
+            { "Bass Boost", std::vector<float>(PRESET_BASS_BOOST.begin(), PRESET_BASS_BOOST.end()) },
+            { "Team Chat", std::vector<float>(PRESET_TEAM_CHAT.begin(), PRESET_TEAM_CHAT.end()) },
+            { "Shooter", std::vector<float>(PRESET_SHOOTER.begin(), PRESET_SHOOTER.end()) },
+            { "MOBA", std::vector<float>(PRESET_MOBA.begin(), PRESET_MOBA.end()) },
+        };
+        return presets;
     }
 
     std::optional<EqualizerInfo> getEqualizerInfo() const override
@@ -81,6 +102,7 @@ public:
         case CAP_BATTERY_STATUS:
         case CAP_SIDETONE:
         case CAP_INACTIVE_TIME:
+        case CAP_EQUALIZER_PRESET:
         case CAP_EQUALIZER:
         case CAP_PARAMETRIC_EQUALIZER:
             return { .usagepage = 0xffa0, .usageid = 0x0001, .interface_id = 3 };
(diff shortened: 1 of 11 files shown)


ChrisLauinger77:
Info was gathered here:
https://github.com/ksanislo/Solaar-NG?tab=readme-ov-file

Implementation was done using GPT model 5.4

@mryll fyi

Should fix #477 for good.
maybe you also wanna adopt the G522 @iowi479 ? the #488 could be closed, too

lib/devices/logitech_gpro_x2_lightspeed.hpp
```diff
@@ -205,33 +441,357 @@ class LogitechGProX2Lightspeed : public HIDDevice {
...
+            2147483648.0
+        };
+
+        std::vector<uint16_t> words;
+        words.reserve(10);
+
+        for (size_t i = 0; i < coeffs.size(); ++i) {
+            auto q_value = static_cast<int64_t>(std::llround(coeffs[i] * scales[i]));
+            q_value      = std::clamp<int64_t>(q_value, -(1LL << 31), (1LL << 31) - 1);
+            q_value &= 0xFFFFFF00;
```
Sapd: Sign loss for negative coefficients here. `0xFFFFFF00` is a positive 32-bit literal that zero-extends to `int64_t` as `0x00000000FFFFFF00`. AND-ing a negative clamped value (e.g. `-100` = `0xFFFFFFFFFFFFFF9C`) with that erases the sign-extension bits and turns it into a large positive number. Biquad `b1`/`a1` are routinely negative (`-2*cos(w0)/a0`), so the encoded coefficients for low/mid bands come out wrong. Use `q_value &= ~INT64_C(0xFF);` instead. Only affects the onboard EQ fallback, but it's still wrong.

lib/devices/logitech_gpro_x2_lightspeed.hpp
```diff
@@ -57,6 +113,21 @@ class LogitechGProX2Lightspeed : public HIDDevice {
...
+                static_cast<uint16_t>(protocols::CenturionFeature::CenturionBatterySoc),
+                0x00);
+            centurion_battery) {
+            auto battery_result = parseCenturionBatteryResponse(*centurion_battery);
+            if (!battery_result) {
+                return battery_result.error();
+            }
+
+            battery_result->raw_data       = *centurion_battery;
+            battery_result->query_duration = std::chrono::milliseconds { 0 };
```
Sapd: `query_duration` is hard-coded to 0 on the Centurion success path while the legacy fallback below measures it. Either measure both or drop the field — silently zero is misleading.

lib/devices/logitech_gpro_x2_lightspeed.hpp
```diff
@@ -191,7 +392,42 @@ class LogitechGProX2Lightspeed : public HIDDevice {
...
+            return DeviceError::protocolError("Empty Centurion battery response");
+        }
+
+        auto level = static_cast<int>(packet[0]);
+        if (level < 0 || level > 100) {
+            return DeviceError::protocolError("Centurion battery percentage out of range");
+        }
+
+        auto charging_state = packet.size() >= 3 ? packet[2] : 0;
+        auto status         = (charging_state == 1 || charging_state == 2)
```
Sapd, line 407: Centurion treats states 1 and 2 as charging, but `parseBatteryResponse` above only treats `0x02` as charging. Probably correct (different encodings), but a one-line comment here would stop someone from "fixing" the inconsistency later.

lib/devices/logitech_gpro_x2_lightspeed.hpp
```diff
@@ -40,7 +46,54 @@ class LogitechGProX2Lightspeed : public HIDDevice {
...
+
+    std::optional<EqualizerInfo> getEqualizerInfo() const override
+    {
+        return EqualizerInfo {
+            .bands_count    = cached_band_count_,
+            .bands_baseline = 0,
+            .bands_step     = 1.0f,
+            .bands_min      = cached_gain_min_,
+            .bands_max      = cached_gain_max_
+        };
```
Sapd, line 84: `cached_band_count_`/`cached_gain_min_`/`cached_gain_max_` start at hard-coded defaults (5/-12/+12) and only get refreshed when `readEqualizerDescriptor` runs. A caller that asks `getEqualizerInfo()` before any other operation gets fabricated values that may not match the device. Worth a doc note or just returning `std::nullopt` until the cache is populated.

lib/devices/logitech_gpro_x2_lightspeed.hpp
```diff
@@ -108,45 +178,176 @@ class LogitechGProX2Lightspeed : public HIDDevice {
...
+            preset_values = &PRESET_TEAM_CHAT;
+            break;
+        case 3:
+            preset_values = &PRESET_SHOOTER;
+            break;
+        case 4:
+            preset_values = &PRESET_MOBA;
+            break;
+        default:
+            return DeviceError::invalidParameter("Device only supports presets 0-4");
```
Sapd: Dead branch — the bounds check above already returns for `preset >= EQUALIZER_PRESETS_COUNT`.

lib/devices/protocols/logitech_centurion_protocol.hpp
```diff
@@ -0,0 +1,433 @@
...
+
+        if (reply_data[5] == 0xFF) {
+            return DeviceError::protocolError("Centurion sub-device rejected the request");
+        }
+
+        return std::vector<uint8_t>(reply_data.begin() + 7, reply_data.end());
+    }
+
+protected:
+private:
```
Sapd, line 199: Empty `protected:` immediately followed by `private:` — drop the `protected:`.

lib/devices/protocols/logitech_centurion_protocol.hpp
```diff
@@ -0,0 +1,433 @@
...
+        std::span<const uint8_t> params = {}) const
+    {
+        auto payload = buildDirectPayload(feature_index, function, params);
+        auto frame   = buildCenturionFrame(payload);
+
+        if (auto write_result = this->writeHID(device_handle, frame, frame.size()); !write_result) {
+            return write_result.error();
+        }
+
+        for (int attempt = 0; attempt < 8; ++attempt) {
```
Sapd: The 8-attempt poll appears here and again in `sendCenturionBridgeRequest`. Pull it out as a named constant so both stay in sync, and so the magic number has a name.

lib/devices/protocols/logitech_centurion_protocol.hpp
```diff
@@ -0,0 +1,433 @@
...
+            return DeviceError::protocolError("Centurion bridge index not initialized");
+        }
+
+        auto sub_message = buildBridgeSubMessageVector(sub_feature_index, function, params);
+        const size_t sub_message_size = sub_message.size();
+
+        std::vector<uint8_t> bridge_prefix {
+            *centurion_bridge_index_,
+            static_cast<uint8_t>(BRIDGE_SEND_FRAGMENT_FN | SOFTWARE_ID),
+            static_cast<uint8_t>((sub_message_size >> 8) & 0x0F),
```
Sapd, line 323: `(sub_message_size >> 8) & 0x0F` caps fragmented messages at 4095 bytes. Current payloads are well under that, but a `static_assert` or runtime check would catch a future regression.

Sapd (commented):
A few things from a read-through.

ChrisLauinger77:
@Sapd changes are implemented (yesterday) - forgot to mention this here

Issue #460, opened 2026-01-20
Audeze Maxwell not reporting any battery status


T13nou:
### Description

Hi ! 

I've tried to use HeadSetControl with my Audeze Maxwell but unfortunately it's not able to report any battery status.

The device is marked as supported and here is what I have when I try to read the battery level 

`❯ headsetcontrol -b
Found 1 supported device(s):
 Audeze Maxwell (Audeze Maxwell XBOX Dongle) [0x3329:0x4b18]
Battery:
        Status: BATTERY_UNAVAILABLE
`

Thanks for helping !

T13nou

### Headset Name

Audeze Maxwell

### On which OS does the problem happen?

Linux

### Device information

<details>
 <summary>Detailed Device Information</summary>

 ```
 ❯ headsetcontrol --dev -- --list --device 0x3329:0x4b18
Device Found
  VendorID: 0x3329
 ProductID: 0x4b18
  path: /dev/hidraw8
  serial_number: 0000000000000000
  Manufacturer: Audeze LLC
  Product:      Audeze Maxwell XBOX Dongle
  Interface:    0
  Usage-Page: 0xc Usageid: 0x1

Device Found
  VendorID: 0x3329
 ProductID: 0x4b18
  path: /dev/hidraw8
  serial_number: 0000000000000000
  Manufacturer: Audeze LLC
  Product:      Audeze Maxwell XBOX Dongle
  Interface:    0
  Usage-Page: 0xff13 Usageid: 0x1

Device Found
  VendorID: 0x3329
 ProductID: 0x4b18
  path: /dev/hidraw8
  serial_number: 0000000000000000
  Manufacturer: Audeze LLC
  Product:      Audeze Maxwell XBOX Dongle
  Interface:    0
  Usage-Page: 0xb Usageid: 0x5

 ```

</details>

ModoloDev:
The "standard" version works on Windows.

```
>headsetcontrol-windows-x86_64.exe -b
Found 1 supported device(s):
 Audeze Maxwell (Audeze Maxwell HID) [0x3329:0x4b19]
Battery:
        Status: BATTERY_AVAILABLE
        Level: 60%
```

Have you tried it more than once?

T13nou:
> The "standard" version works on Windows.
> 
> ```
> >headsetcontrol-windows-x86_64.exe -b
> Found 1 supported device(s):
>  Audeze Maxwell (Audeze Maxwell HID) [0x3329:0x4b19]
> Battery:
[...]

PR #517, opened 2026-05-06
Audeze Maxwell Noise Filter Capability and Audeze Maxwell 2 Protocol

ModoloDev:
### Changes made
- Added noise filter capability
- Replicated Audeze Maxwell 2 15-packet initialization

Related to #460, it might fix some of the problems with battery status

### Checklist

### DIFF
--- a/lib/devices/audeze_maxwell.hpp
+++ b/lib/devices/audeze_maxwell.hpp
@@ -22,10 +22,11 @@ namespace headsetcontrol {
  * - Battery status
  * - Chatmix
  * - Voice prompts
+ * - Mic noise filter
  *
  * Special Requirements:
- * - Complex 18-packet initialization sequence
- * - 60ms delay between packets (mimics official software)
+ * - 15-packet initialization sequence followed by 6 status requests
+ * - 60ms delay between packets (official software sends at 52ms but 60ms is more reliable in testing)
  * - Uses hid_get_input_report for responses
  */
 class AudezeMaxwell : public HIDDevice {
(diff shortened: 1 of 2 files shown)


Nallebeorn:
I'm running a build of this branch (on Fedora 44 KDE) and it's working well for me so far. Compared to master it seems to be much more reliable about successfully reporting battery status immediately after connecting. I have the PlayStation edition of the original Maxwell (not 2).

It can still be a bit wonky and imprecise, with the battery level going up and down by a few points randomly. One time it reported 0% on the first run after connecting. But like I said in #460 it's not obviously any worse than the official app, and could just be down to how the battery level is measured in the first place like Sapd said (this isn't a topic I really know anything about).

lib/devices/audeze_maxwell.hpp
Sapd: Just noticed for some reason the array size is wrong, we could directly fix it here

lib/devices/audeze_maxwell.hpp
```diff
@@ -236,29 +261,11 @@ class AudezeMaxwell : public HIDDevice {
...
-                final_minutes = 45;
-            else if (minutes <= 60)
-                final_minutes = 60;
-            else if (minutes <= 90)
-                final_minutes = 90;
-            else if (minutes <= 120)
-                final_minutes = 120;
-            else if (minutes <= 240)
-                final_minutes = 240;
+            uint16_t final_minutes = map(minutes, 0, 90, 0, 360);
```
Sapd: Seems to be semantically incorrect.

Passing in 5 minutes, the user would get 20

lib/devices/audeze_maxwell.hpp
```diff
@@ -129,64 +131,87 @@ class AudezeMaxwell : public HIDDevice {
...
             }
-
-            // Check microphone status (note: BatteryResult has mic_status field)
-            if (status_buffs[1][12] == 0xFF) {
-                status.battery.mic_status = MICROPHONE_UP;
-            }
         }
 
+        // Parse microphone mute status (FF = unmuted)
+        status.mic_muted = status_buffs[1][12] != 0xFF;
```
Sapd: We could in pass that information btw in BatteryResult back

lib/devices/audeze_maxwell.hpp
```diff
@@ -218,7 +243,7 @@ class AudezeMaxwell : public HIDDevice {
 
         if (mapped == 0) {
             // Toggle sidetone off
-            std::array<uint8_t, MSG_SIZE> toggle_cmd { 0x6, 0x9, 0x80, 0x5, 0x5a, 0x5, 0x0, 0x82, 0x2c, 0x7, 0x0, 0x1 };
+            std::array<uint8_t, MSG_SIZE> toggle_cmd { 0x06, 0x09, 0x80, 0x05, 0x5A, 0x05, 0x00, 0x82, 0x2C, 0x07, 0x00, 0x00 };
```
Sapd, line 247: Was it before the other way around? As its now 0x00 as last byte

PR #527, opened 2026-05-26
Added support for Corsair Virtuoso XT

Anshuman01000001:
Adds support for the Corsair Virtuoso XT — both the wireless receiver (0x0a64) and wired USB (0x0a62).

I reverse-engineered the protocol via hidraw probing on Linux. The Virtuoso XT doesn't use the same protocol as CorsairVoidRich or CorsairVoidV2W — it communicates on interface 3 (Usage-Page 0xff42) and responds to a simple 0x02 0x00 battery request:

- Byte 0: report ID (0x01)
- Byte 1: status (0xf0 = connected, 0x00 = offline)
- Byte 2: battery percentage (0-100)

Charging state doesn't appear to be reported over HID. Tested on Arch Linux with the wireless receiver — battery percentage matches the headset's LED indicator.

Capabilities: battery status only for now.

### DIFF
--- a/lib/device_registry.cpp
+++ b/lib/device_registry.cpp
@@ -3,6 +3,7 @@
 // Corsair devices
 #include "devices/corsair_void_rich.hpp"
 #include "devices/corsair_void_v2w.hpp"
+#include "devices/corsair_virtuoso_xt.hpp"
 
 // Logitech devices with HIDPPDevice protocol template
 #include "devices/logitech_g432.hpp"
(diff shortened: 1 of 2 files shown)


Sapd:
Ignore the review about the comments. However the first bug might be relevant.

Anshuman01000001:
I'll make the required changes. I will also try and adding more features for this headest after resolving this PR, Thanks.

PR #537, opened 2026-07-03
Add HS80 Max support

emericg:
### Changes made

Added HS80 Max (0xa97) to the CorsairVoidV2W class. Sidetone confirmed working.
I did not add 0x2a0f ID because another issue mentioned it was a different type of HS80 Max?

The battery level wasn't working, as stated in the previous comment all I got was 271%, the 0x0a96 headset ID, but the battery level always follows in the next packet or so, so a simple retry was enough. Maybe that's made to signal per device battery level, but I only have the one to validate that theory.

I added a wakeDevice() function that doesn't do the full "software mode" switch, because it turns out that was what was producing audible pop sound + subtle volume level change in my headset, made worse by using the HeadsetControl applet for Gnome (a query every couple of minutes).

The pop still happens when changing sidetone or setting the inactive timer, but it's more acceptable because it's triggered following a user action and not happening out of nowhere.

Implemented setting the inactive timer and MICROPHONE_UP support (both taken from OpenLinkHub code).

Cleanup the code style a bit, in a separate commit if you don't want it, I would personally also rename the class to get rid of the "void" in the title.
And thinking about it now, I did not changed the README to add "inactive time" and "rotate to mute" (but I only tested with my HS80 Max).

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ A cross-platform tool to control USB gaming headsets on **Linux**, **macOS**, an
 | Logitech Zone Wired/Zone 750 | All | x |   |   |   |   |   | x | x |   |   |   |   |   |   |   |   |   |
 | Logitech ASTRO A50 Gen 5 | All | x | x |   | x |   | x |   |   |   |   | x |   |   |   |   |   | x |
 | Corsair Headset Device | All | x | x | x | x |   |   |   |   |   |   |   |   |   |   |   |   |   |
-| Corsair Wireless V2 Headset Device | All | x | x |   |   |   |   |   |   |   |   |   |   |   |   |   |   |   |
+| Corsair Wireless V2 Headset Device | All | x | x |   |   | x |   |   |   |   |   |   |   |   |   |   |   |   |
 | SteelSeries Arctis (1/7X/7P) Wireless | All | x | x |   |   | x |   |   |   |   |   |   |   |   |   |   |   |   |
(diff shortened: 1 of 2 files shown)


Sapd:
> Cleanup the code style a bit, in a separate commit if you don't want it, I would personally also rename the class to get rid of the "void" in the title.

Feel free to do so

> And thinking about it now, I did not changed the README to add "inactive time" and "rotate to mute" (but I only tested with my HS80 Max).

Best to regenerate it using the --readme-helper parameter

emericg:
Ok so browsing a bit more of the OpenLinkHub code, while they do per device implementations, they also clearly identify the slipstream and slipstream v2 (looking at the Corsair website, it's the commercial name) as the two variants of the protocol, and I guess if not going too deep in device specific features, the two classes of HeastsetControl should handle most of the Corsair headsets.

Following the getDeviceName() which says "Corsair Wireless V2 Headset Device", renaming the class CorsairVoidV2W > CorsairWirelessV2 and the file corsair_void_v2w.hpp > corsair_wireless_v2 looks good to me.

Now that breaks the similarities with corsair_void_rich.hpp, so rename one, both, none? Your call honestly, it's not that important.

Also unrelated, I noticed a stray logitech.h file that survived the C++ modernization :D I can include it in that pull request if you'd like.

humplemann:
@emericg Thanks! Very interested to see if that finds its way into a release (and when). Again thanks.

PR #549, opened 2026-07-29
perf(cli): skip the forced info reads when the invocation performs an action

AdrianKuriata:
Structured output (json/yaml/env) marks every info capability as requested, no
matter what the command line actually asked for. On a device with an expensive
status read that dominates an action which is otherwise instant.

Audeze Maxwell 2 (0x3329:0x4b28), release build:

    headsetcontrol -d 0x3329:0x4b28 -s 20                 0.07 s
    headsetcontrol -d 0x3329:0x4b28 -s 20 --output=json   2.90 s

The difference is the status read - getDeviceStatus() sends 15 init packets plus
6 status requests, each after a 60 ms delay. The sidetone write itself is one
packet. Anything driving the CLI per user interaction pays that every time; a GUI
moving a slider ends up queueing writes for minutes.

After the patch the same call takes 0.07 s and the actions array is unchanged. A
pure query is unaffected: --output json still takes 2.83 s and still reports
battery and chatmix. Info asked for explicitly still works, -b -s 20 -o json
reports the battery in 1.49 s.

headsetcontrol_tests passes.

One thing to decide: this changes the json shape for action invocations,
devices[].battery is no longer there when you only asked to set something. I think
the old behaviour is the bug, but if you would rather keep the default I can put
it behind a flag instead.

### DIFF
--- a/cli/main.cpp
+++ b/cli/main.cpp
@@ -1031,14 +1031,13 @@ std::vector<DeviceList> toLegacyDeviceList(std::vector<DiscoveredDevice>& device
     return legacy;
 }
 
-// Whether this invocation was asked to set something.
-bool hasRequestedAction(const std::vector<DiscoveredDevice>& devices)
+// Whether this device was asked to set something. Only counts requests that are
+// still live, so it has to run after handleMultiDeviceActions has had its say.
+bool hasRequestedAction(const DiscoveredDevice& device)
 {
-    for (const auto& dev : devices) {
-        for (const auto& req : dev.feature_requests) {
-            if (req.type == CAPABILITYTYPE_ACTION && req.should_process) {
-                return true;
-            }
+    for (const auto& req : device.feature_requests) {
+        if (req.type == CAPABILITYTYPE_ACTION && req.should_process) {
+            return true;
         }
     }
 
@@ -1047,13 +1046,17 @@ bool hasRequestedAction(const std::vector<DiscoveredDevice>& devices)
 
 // Enable info requests for extended output formats (JSON, YAML, ENV)
 // Only for a pure query: the extra reads can cost far more than the action
-// itself (Maxwell 2: -s 20 is 0.07 s, -s 20 -o json is 2.90 s).
+// itself (Maxwell 2: -s 20 is 0.07 s, -s 20 -o json is 2.90 s). Per device, so
+// an action on one headset does not silence the info reads of another.
 void enableExtendedInfoRequests(std::vector<DiscoveredDevice>& devices, bool extended)
 {
-    if (!extended || hasRequestedAction(devices))
+    if (!extended)
         return;
 
     for (auto& dev : devices) {
+        if (hasRequestedAction(dev))
+            continue;
+
         for (auto& req : dev.feature_requests) {
             if (req.type == CAPABILITYTYPE_INFO && !req.should_process && dev.hasCapability(req.cap)) {
                 req.should_process = true;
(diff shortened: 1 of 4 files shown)


Sapd:
Thanks, the numbers speak for themselves here. I agree the old behaviour is the bug, so no flag please, keep it as the default.

Two things before this can go in.

The output shape changes, so please bump API_VERSION in cli/output/output.cpp to 1.5. Thats what it is there for, otherwise a consumer has no way to notice.

And enableExtendedInfoRequests runs before handleMultiDeviceActions, so hasRequestedAction looks at action requests that get neutralized one line later. It also checks all devices at once, so one device with an action switches off the info reads of another device that never had one. Please move the call after handleMultiDeviceActions and do the check per device instead of globally.

Also worth writing into the PR text: it does not only drop info when you set something, it drops everything you did not explicitly ask for. With -b -s 20 -o json you now get battery but no chatmix, master returned both. Fine by me, but people will notice.

Not caused by you, I just saw it while reading this: the multiple devices guard in handleMultiDeviceActions does not actually do anything. It sets result.message and value but leaves should_process true, and status was already NotProcessed, so processFeatureRequests runs the action anyway. I will fix that separately. It is also why the ordering above matters, once that guard works your version would return neither the action nor the info.

AdrianKuriata:
All three done in b899244.

`handleMultiDeviceActions` now runs first and `enableExtendedInfoRequests` after it, so the check sees action requests as they actually stand rather than as they were parsed. I left a comment at the call site saying why the order is what it is, otherwise it is the kind of thing that gets swapped back.

`hasRequestedAction` takes a single device now and the caller skips per device, so an action on one headset no longer silences the info reads of another.

`API_VERSION` is 1.5. I also updated the sample response in `docs/LIBRARY_USAGE.md`, it had the old number in it.

On the shape change you asked me to write down: it is broader than the title suggests, an invocation that sets something now reports only the info you named. `-b -s 20 -o json` gives you battery and no chatmix where master gave both. I have not edited the PR description, so this comment is where that note lives.

Checked on the test device: `-s 20 -o json` returns the actions array and no battery or chatmix, `-o json` still returns both, `-b -s 20 -o json` returns battery only. With a filter and two devices the non-matching device is not emitted at all, so the per-device check does not leak unprocessed info entries into the output. `make check` passes, 2/2.

Two caveats. I could not run clang-format, there is no clang-format 18 on this machine, so formatting is by hand and CI is the first real check on it. And the broken multi-device guard is the one path I cannot exercise here, since I only have one test device plus one real headset. Once your separate fix lands and `should_process` actually goes false, this version enables the info reads for that device, which I believe is what you wanted, but I have not seen it run.

AdrianKuriata:
Correction to my last comment: I said I had no clang-format 18. I got 18.1.3 afterwards, the same version CI installs, and ran it. cpp-linter passes now.

One extra commit, 43a4a07, and it is not mine in substance. The workflow runs with `lines-changed-only: false`, so it checks the whole of every file a PR touches, and both files this PR touches were already non-conforming: `cli/main.cpp` at lines 37, 423 and 1006, `cli/output/output.cpp` at line 184. None of them are in this PR's diff. The commit is clang-format 18.1.3 output, nothing hand-written, and I kept it separate so you can see the review fixes and the formatting apart. Say the word if you would rather have it out and handled on its own.

Sapd:
Pushed 9a438ac directly on top of your branch: added three CLI-level tests covering the path this PR actually changes. Every existing JSON test passed `-b` alongside the action flag, so `should_process` for battery was already true before `enableExtendedInfoRequests` ran and the new suppression logic never got exercised. New cases: action-only (`-s 20 -o json`) omits battery/chatmix, action + explicit info (`-b -s 20 -o json`) keeps only battery, and a pure query still returns both. All pass against the current branch.

PR #556, opened 2026-08-02
fix: enable Windows support for Arctis Nova 3P/3X via output reports

RpYTMc:
### Changes made

Enables Windows support for the SteelSeries Arctis Nova 3P/3X Wireless by switching all device communication from HID **feature reports** to **output reports**, matching the existing Nova 5/7 implementations.

**Why this fixes Windows:** on Windows, hidapi pads feature reports to the interface's declared feature-report length (~1060 bytes on this device), and the firmware silently ignores oversized reports — so every feature-report command was a no-op there. This is the "commands have no effect on Windows" problem observed in #417 when the device was added, which led to the `PLATFORM_LINUX | PLATFORM_MACOS` restriction. Output reports don't hit this issue, so the same packets that already work for the Nova 5/7 work here too.

Details:

- `getBattery()` now uses the shared `readDeviceStatus()` helper (`[0x00, 0xb0]` write + read), like Nova 5/7. The status response layout is identical to the Nova 5: `data[1] == 0x02` → offline, `data[4] == 0x01` → charging, `data[3]` → level percent.
- All commands (sidetone, mic volume, inactive time, equalizer, parametric equalizer, save state) now go through `writeHID()` with the `0x00` report-ID prefix; EQ payload offsets shift by one byte accordingly.
- Adds **charging detection**, which the previous implementation didn't report.
- Removes the `getSupportedPlatforms()` override (base default is `PLATFORM_ALL`) and updates the README row from `L/M` to `All`.

**Tested on real hardware** — Arctis Nova 3X (`0x1038:0x226d`), Windows 11, MSYS2/MinGW64 build:

- Battery level reads correctly (matches an independent node-hid implementation of the same query)
- Charging flag flips when plugging/unplugging the headset
- Sidetone audibly changes across levels
- Equalizer presets audibly change (Flat ↔ Smiley)
- Inactive time accepted without timeout
- Unit tests pass (`ctest`: 2/2)

The byte meanings (`data[4]`: `0x01` charging / `0x03` discharging; `data[1]`: `0x03` online / `0x02` off) were confirmed by live HID captures while docking/undocking the headset.

Note: I could only test on Windows. The write path used here is byte-identical to what the Nova 5/7 already use successfully on Linux/macOS, but a quick Linux/macOS smoke test from another 3P/3X owner would be welcome.

### Checklist

- [ ] For new features in HeadsetControl: I discussed it beforehand in Issues or Discussions and adhered to the [wiki](https://github.com/Sapd/HeadsetControl/wiki/Development#adding-a-new-feature-to-the-headsestcontrol-application-itself) — N/A, this fixes existing device support rather than adding a feature

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,7 +176,7 @@ sudo udevadm control --reload-rules && sudo udevadm trigger
 | SteelSeries Arctis Nova 7P | All |   | x |   |   | x |   |   |   | x | x |   | x | x | x | x | x |   |
 | SteelSeries Arctis 7+ | All | x | x |   |   | x | x |   |   | x | x |   |   |   |   |   |   |   |
 | SteelSeries Arctis Nova Pro Wireless | All | x | x |   | x | x |   |   |   | x | x |   |   |   |   |   |   |   |
-| SteelSeries Arctis Nova 3P Wireless | L/M | x | x |   |   | x |   |   |   | x | x | x |   | x |   |   |   |   |
+| SteelSeries Arctis Nova 3P Wireless | All | x | x |   |   | x |   |   |   | x | x | x |   | x |   |   |   |   |
 | SteelSeries Arctis Buds | All |  | x |   |   |  |   |   |   |  |  |  |   |  |   |   |   |   |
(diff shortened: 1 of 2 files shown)


Sapd:
fine because I think a smoke test will only come once merged

PR #571, opened 2026-09-05
Fix -d device selection silently matching nothing for bare hex USB IDs

gdevenyi:
### Changes made

`-d 1b1c:0a64` selects nothing:

```console
$ headsetcontrol -b
 Corsair Virtuoso XT/SE (CORSAIR VIRTUOSO XT Wireless Gaming Receiver) [0x1b1c:0x0a64]
Battery:
	Level: 59%

$ headsetcontrol -d 1b1c:0a64 -b
No supported device found
```

`parse_two_ids()` reads a token without a `0x` prefix as decimal, and `std::from_chars` reports success when it consumes only *part* of the input. `"1b1c"` therefore stops at the `b` and yields `1`, `"0a64"` yields `0`. The filter then looks for device `0001:0000`, matches nothing, and the failure surfaces as "No supported device found" — as though the device were unplugged.

The help text has advertised exactly this form all along:

```
  -d, --device VID:PID              Select device (e.g., 1038:12ad)
```

`12ad` cannot be anything but hex, so the documented example could never have worked.

**Fix.** `parse_two_ids()` gains a base for tokens without a `0x` prefix, defaulting to 10 so existing callers are unaffected. The two call sites that parse USB vendor/product IDs — `-d` in `main.cpp` and `--device` in `dev.cpp` — ask for base 16, which is how `lsusb` and this tool's own device listing print them. An explicit `0x` prefix still forces hex regardless of the base, so `-d 0x1b1c:0x0a64` keeps working.

`--usage` in dev mode is *not* a USB ID and its help documents a decimal range (`USAGEPAGE:USAGEID (0-65535)`), so it keeps the decimal default — that's why this is a per-call-site base rather than a global switch.

Partial parses are also rejected outright now instead of being truncated, so a genuinely malformed value reports a format error rather than quietly becoming a filter that matches nothing.

**Range checking** (added after review). `-d` cast the parsed IDs straight to `uint16_t`. An out-of-range value wraps to `0`, and `matchesDevice()` treats `0` as "no filter", so `-d 10000:0a64` matched a device from *any* vendor and `-d 10000:10000` disabled filtering entirely and matched everything connected. Both are now rejected. `0` itself stays accepted, since that is the existing "no filter" default and `-d 1b1c:0` is a legitimate way to select any product from one vendor. Dev mode already range-checked its own `--device`.

### Behaviour change to be aware of

`-d 1234:5678` previously meant decimal `(1234, 5678)` and now means `(0x1234, 0x5678)`. I think that's the correct reading for a USB ID — decimal USB IDs aren't a thing anyone writes, and the flag's own help documents hex — but it is a change, so flagging it explicitly. If you'd rather not touch it, the conservative subset of this PR is just the partial-parse rejection plus the range check, which turns the silent failures into clear errors without reinterpreting anything; happy to trim it down to that.

### Testing

```console
$ headsetcontrol -d 1b1c:0a64 -b        # was: No supported device found
	Level: 59%
$ headsetcontrol -d 0x1b1c:0x0a64 -b    # unchanged
	Level: 59%
$ headsetcontrol -d 1038:12ad -b        # device not present, correctly filtered
No supported device found
$ headsetcontrol -d ffff:ffff -b        # in range, just not present
No supported device found
$ headsetcontrol -d zz:yy -b            # was: No supported device found
Error: device: format: vendorid:productid
$ headsetcontrol -d 10000:0a64 -b       # was: matched a Corsair device
Error: device: ids must be between 0 and ffff
$ headsetcontrol -d 10000:10000 -b      # was: matched everything
Error: device: ids must be between 0 and ffff
$ headsetcontrol --dev -- --device 1b1c:0a64 --list
Device Found
  VendorID:  0x1b1c
  ProductID: 0x0a64
  ...
```

Extended `testParseTwoIds` to cover bare hex in base 16, the `0x` prefix overriding the requested base, the same digits parsing differently in base 10 vs 16, and rejection of partly-numeric and non-numeric tokens. Existing assertions — including `"1234:5678"` → `(1234, 5678)` in the default base — are unchanged and still pass. Full unit suite green, `clang-format` 18 clean.

### Checklist

Bug fix rather than a new feature, and no README change needed — the documented syntax now behaves as documented.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01BHCVMmAAeV5WGP7oYZsk3T

### DIFF
--- a/cli/dev.cpp
+++ b/cli/dev.cpp
@@ -212,10 +212,10 @@ struct DevOptions {
     for (int c; (c = getopt_long(argc, argv, "d:i:lu:s:m:f:rg:t:hR:", long_opts.data(), &option_index)) != -1;) {
         switch (c) {
         case 'd': {
-            auto ids = headsetcontrol::parse_two_ids(optarg);
+            auto ids = headsetcontrol::parse_two_ids(optarg, 16);
             if (!ids || !in_range(ids->first, 1, 65535) || !in_range(ids->second, 1, 65535)) {
-                std::cerr << "Invalid --device. Use format: VENDORID:PRODUCTID (1-65535 or 0x1-0xffff)\n"
-                          << "  Example: --device 0x1b1c:0x1b27\n";
+                std::cerr << "Invalid --device. Use format: VENDORID:PRODUCTID (hex, 1-ffff)\n"
+                          << "  Example: --device 1b1c:1b27\n";
                 return std::nullopt;
             }
             opts.vendorid  = static_cast<uint16_t>(ids->first);
--- a/cli/main.cpp
+++ b/cli/main.cpp
@@ -191,9 +191,18 @@ std::optional<cli::ParseError> configureParser(cli::ArgumentParser& parser, Opti
         .custom('d', "device", cli::ArgRequirement::Required, [&opts](std::optional<std::string_view> arg) -> std::optional<cli::ParseError> {
                 if (!arg)
                     return cli::ParseError { "requires vendor:product", "device" };
-                auto ids = headsetcontrol::parse_two_ids(*arg);
+                // USB IDs are hex, and are written without a 0x prefix by lsusb
+                // and by our own device listing.
+                auto ids = headsetcontrol::parse_two_ids(*arg, 16);
                 if (!ids) {
                     return cli::ParseError { "format: vendorid:productid", "device" };
+                }
+                // Narrowing an out-of-range ID would wrap it, and a wrapped-to-zero
+                // ID reads as "no filter" - so -d 10000:10000 would silently match
+                // every device instead of none.
+                constexpr int id_max = 0xffff;
+                if (ids->first < 0 || ids->first > id_max || ids->second < 0 || ids->second > id_max) {
+                    return cli::ParseError { "ids must be between 0 and ffff", "device" };
                 }
                 opts.vendor_id = static_cast<uint16_t>(ids->first);
                 opts.product_id = static_cast<uint16_t>(ids->second);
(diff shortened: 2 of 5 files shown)


gdevenyi:
Addressed the range-check point from the review — it was right, and it reproduces:

```console
$ headsetcontrol -d 10000:0a64 -b     # asked for vendor 0x10000
 Corsair Virtuoso XT/SE (CORSAIR VIRTUOSO XT Wireless Gaming Receiver) [0x1b1c:0x0a64]
	Level: 59%

$ headsetcontrol -d 10000:10000 -b    # asked for a vendor and product that cannot exist
 Corsair Virtuoso XT/SE (CORSAIR VIRTUOSO XT Wireless Gaming Receiver) [0x1b1c:0x0a64]
	Level: 59%
```

`0x10000` wraps to `0` on the narrowing cast, and `matchesDevice()` treats `0` as "no filter", so the first case matches a device from any vendor and the second disables filtering completely and matches everything. Worth noting it needs a *matching* product to show up — `-d 10000:1` still prints "No supported device found" because the product filter rejects it, which is probably why it went unnoticed.

That's the same failure mode this PR exists to fix — a malformed `-d` silently selecting the wrong thing instead of reporting bad input — so it belongs here rather than in a follow-up. `-d` now range-checks both IDs before narrowing:

```console
$ headsetcontrol -d 10000:0a64 -b
Error: device: ids must be between 0 and ffff
$ headsetcontrol -d -1:5 -b
Error: device: ids must be between 0 and ffff
$ headsetcontrol -d ffff:ffff -b      # still valid, just not present
No supported device found
$ headsetcontrol -d 1b1c:0a64 -b      # unchanged
	Level: 59%
```

`0` stays accepted, since that is the existing "no filter" default and `-d 1b1c:0` is a legitimate way to select any product from one vendor. Dev mode already range-checked its own `--device` via `in_range()`, so only `main.cpp` needed it.

PR #583, opened 2026-09-19
Arctis 7+: persist settings with the save command

steveyh25:
On the Arctis 7+ family, sidetone, inactive time and equalizer changes were lost whenever the headset was power-cycled, because the driver never sent a save. This adds a saveStateArctis7Plus() helper that sends 00 09 (the same save opcode and pattern steelseries_arctis_nova_3p_wireless.hpp already uses) - called in each of the four setters.

Tested on an Arctis 7X+ (1038:2216): sidetone, EQ preset, custom EQ and inactive time all survive a power-cycle with this change. The other 7+ product IDs (220e, 2212, 2236) are untested, but they are the same hardware generation so I suspect the fix is identical / works on all. Each set now writes to flash (which may matter to anyone calling these from a timer, as I was doing previously). There already appears to be precedent in the code that save-after-set is an approved idiom, however.


Sapd (commented):
Matches what steelseries_arctis_nova_3p_wireless.hpp already does - same 00 09, same helper shape, same call pattern. Builds clean, tests pass.

One thing: put your note about every set writing flash into the header comment, not just the PR. Anyone calling these on a timer needs to see it in the file.

The untested 7+ PIDs are fine by me, same generation.

